PR #169, opened 2016-03-10
fix some spell/grammatical errors

PARAG00991:


tturocy:
These are third-party files, so we will leave them as they are upstream.

PR #228, opened 2017-12-24
Fix error condition in integer parser

mmcco:
The current condition will always be false (`c` can't be equal to both `'+'` and `'-'`). It seems pretty clear that the following was intended.

Thanks for your time,
Michael McConville
University of Utah


tturocy:
I wanted to acknowledge this with thanks.  I will need to merge this to earlier branches in the first instance instead, as this obviously affects several versions - I have to leave that to one side for right now due to a queue of other work but will do so in due course.

Issue #203, opened 2016-11-15
Support for Python3?


lampda:
I downloaded gambit-15.1 from github, and installed to my python site-packages folder.
However, it seems that I can ONLY install gambit with Python2.7, but Python3.5/Python3.6 will throw error that saying "__dict__ is not writable".

robert-7:
I'm not 100% I can help with this, but could you please give the full error message you're getting when trying to install?

tturocy:
This is correct - Python 3 is not supported at present (but we would always welcome patches to add such support!)

fairsky66:
@tturocy Thank you for your confirmation.
@robert-7 I tried to produce the exact error message I had encountered with Python3 in my virtual env, however I could not make it install while came to your comment&request.

Tehnrry:
How to install the gambit to the  python site-packages folder.When I use python setup.py install, encountering some errors. (Python2.7 on win10)
--------------------------------------
E:\gambit-15.1.0\src\python>python setup.py install
running install
running bdist_egg
running egg_info
writing gambit.egg-info\PKG-INFO
writing top-level names to gambit.egg-info\top_level.txt
writing dependency_links to gambit.egg-info\dependency_links.txt
writing entry points to gambit.egg-info\entry_points.txt
reading manifest file 'gambit.egg-info\SOURCES.txt'
writing manifest file 'gambit.egg-info\SOURCES.txt'
installing library code to build\bdist.win-amd64\egg
running install_lib
running build_py
running build_ext
skipping 'gambit/lib\libgambit.cpp' Cython extension (up-to-date)
building 'gambit.lib.libgambit' extension
C:\Users\Tehnrry\AppData\Local\Programs\Common\Microsoft\Visual C++ for Python\9.0\VC\Bin\amd64\cl.exe /c /nologo /Ox /MD /W3 /GS- /DNDEBUG -I../.. -I.. "-IC:\Program Files (x86)\Python\Python27\include" "-IC:\Program Files (x86)\Python\Python27\PC" /Tpgambit/lib\libgambit.cpp /Fobuild\temp.win-amd64-2.7\Release\gambit/lib\libgambit.obj
libgambit.cpp
C:\Users\Tehnrry\AppData\Local\Programs\Common\Microsoft\Visual C++ for Python\9.0\VC\Include\xlocale(342) : warning C4530: C++ exception handler used, but unwind semantics are not enabled. Specify /EHsc
[...]

PR #242, opened 2018-09-07
Ongoing attempt at Python 3 support (progress towards #203)

rhalbersma:
**Introduction**
These 6 commits hopefully prepare the ground for a full Python 2 / 3 compatible code base. 

**Rationale**
The commits do the following:
1. I  have set up [Travis CI](https://travis-ci.org/rhalbersma/gambit) and [Codecov](https://codecov.io/gh/rhalbersma/gambit/branch/future) on my fork in the `future` branch.
2. I followed the `futurize` best practices by running the automated `--stage1` and `--stage2` steps, keeping all tests green. 
3. `futurize` does not touch Cython files, so `functools.reduce` needs to be manually imported.
4. Running the nosetests with `python2 -3` signalled quite a few deprecated warnings concerning user-defined `__eq__` functions without an accompanying `__hash__` function. I have added hash support consistent with equality: if equality is on a tuple/list of class members, then the hashing will be on the same data structure (tuple or list of the data members of `self`).
5. Running `python3 setup.py build` because of the new Python 3 class system (see this [Q&A on StackOverflow](https://stackoverflow.com/questions/40367569/dictionary-not-writable-generic-old-setup-py)). 

**Status**
- The project builds correctly under python2 and runs all unit tests without error.
- Running the unit tests with `python2 -3 nosetests` gives no warnings in code from `gambit`, and only a handful about system code. 
- Running `python3 setup.py build` also builds without errors.
- Running `nosetests` in Python3 mode fails on all unit tests, mostly because of `byte` / `str` incompatibility.

**Guidance needed**
- How would you like to tackle Python2 / Python3 string compatibility? My preferred option would be to treat the current `cython` interface of talking to the C++ code through `std::string` or `char *`. This will keep callers using Python2 happy, but requires converting Python3 unicode string literals to `byte`, e.g. using a `.encode('utf-8')` member function call.
- Gambit also produces quite a bit of `char *` strings from `c_str()` member functions. When those talk to users, then they need to be marshalled through `.decode()`. Separating the internal `c_str()` calls from the ones bubbling up to the interface seems quite a bit of work.
- There are also bound to be some issues with `file()` and `open()` incompatibilities between Python2 and Python3. The way forward is to use `open()` everywhere, with the appropriate `b` flags for byte stream reading. 

If you agree with the approach taken so far, I can try and plough ahead with the string and file issues. Merging the pull request would eventually be best done in one squashed commit, or piece meal, whatever your preference is.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,3 +31,5 @@ gambit.wixobj
 gambit.wixpdb
 installer
 config.h
+stamp-h1
+.coverage
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ python:
 
 before_install:
 # Preinstalled pip packages: nose setuptools wheel
-  - pip install codecov coverage cython
+  - pip install codecov coverage cython future
 
 install:
 # Building from git repository
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,36 @@
+branches:
+  only:
+    - future
+os: linux
+dist: trusty
+language: python
+python:
+  - "2.7"
+
+before_install:
+# Preinstalled pip packages: nose setuptools wheel
+  - pip install codecov coverage cython
+
+install:
+# Building from git repository
+  - aclocal
+  - libtoolize
+  - automake --add-missing
+  - autoconf
+  - ./configure --disable-gui
+  - make
+  - sudo make install
+# Building the Python extensions
+  - cd src/python
+  - python setup.py build
+  - sudo env "PATH=$PATH" python setup.py install
+
+script:
+  - cd gambit/tests
+  - nosetests --with-coverage
+
+after_success:
+  - codecov
+
+notifications:
+  email: false
--- a/contrib/demos/poker.py
+++ b/contrib/demos/poker.py
@@ -1,3 +1,4 @@
+from __future__ import print_function
 import gambit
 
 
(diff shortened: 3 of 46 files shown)


rhalbersma:
@tturocy 4 weeks, no feedback so far, how can I improve this PR?

tturocy:
Many thanks for taking this on.  At present I'm on a secondment to a very heavy admin role, which leaves me no time or mindspace for computational game theory.  So while I am looking forward to having a closer look at this, I don't quite know just when that might be.

rhalbersma:
I have decided to abandon this effort. The remaining obstacles are a bit beyond my Python skills. As a Gambit user, it is considerably less work for me to continue using the existing Gambit Python 2 API and futurize my own user code scripts. 

I'll leave this pull request open so that any future attempt to port to Python 3 can use whatever is deemed necessary.

tturocy:
Your efforts on this are appreciated.  With the announcement that many of the main scientific python stack packages (like NumPy and pandas) are going to drop Python 2 support for new releases from the start of 2019, this is bubbling up higher in my priority list.  However, it's tied up with a number of other packaging issues across the whole project which probably have to be looked at holistically.

rhalbersma:
Python 2 itself will no longer be maintained in 12 months time: https://pythonclock.org/

tturocy:
b0f989d3 now introduces Python 3.7.2 support in the gambit package.

PR #284, opened 2022-01-12
Update and rename INSTALL and README to Markdown files (.md)

EwoutH:
Formats the INSTALL file as a markdown file, which increases readability.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -1,15 +1,14 @@
-General information
--------------------
+### General information
 
 Gambit uses the standard autotools mechanism for configuring and building.
-This should be familiar to most users of Un*ces and MacOS X.  In general,
+This should be familiar to most users of Un\*ces and MacOS X.  In general,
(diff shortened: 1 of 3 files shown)


tturocy:
This is a good idea.  But I think we need at least to add the foreign option for automake, otherwise it will start complaining about INSTALL being missing.

And if we're going to do this, should also consider doing README -> README.md at the same time for consistency?

tturocy:
The AM_INIT_AUTOMAKE options are just separated by a space, not a comma.

If you remove the comma between subdir-objects and foreign, and then do a rebase/squash into one commit, I'll gladly merge.

EwoutH:
Good catch, fixed it.

I don’t have a computer at hand, but feel free to Squash and merge via the GitHub interface. Otherwise I can clean it up tomorrow.

EwoutH:
Rebased, reset and recommitted!

PR #286, opened 2022-01-14
Python maintenance: Black formatting, F-strings and Python 3.6+ features

EwoutH:
This PR contains maintenance of the Python code:

- In the first commit, all Python files are formatted with [Black](https://github.com/psf/black). Consistent formatting improved readability.
- In the second commit, all strings are converted to F-strings with [Flynt](https://github.com/ikamensh/flynt). 12 f-strings were created in 6 files. F-strings are in general more readable, concise and performant.
- In the third commit, the Python code is updated with Python 3.6+ features, using [Pyupgrade](https://github.com/asottile/pyupgrade).
  - utf-8 encoding is now the default ([PEP 3120](https://www.python.org/dev/peps/pep-3120/), Python 3.0+)
  - Replace list comprehensions by Generator Expressions ([PEP 289](https://www.python.org/dev/peps/pep-0289/), Python 2.4+)
  - Replace `yield` loop by `yield from` ([PEP 380](https://www.python.org/dev/peps/pep-0380/), Python 3.3+)
  - Remove the default subclass `(object)` when defining a class

@tturocy This PR contains a lot of (small) code changes, so rebasing it is non-trivial. Could you review it before making too much changes to the Python code base?


tturocy:
Many thanks for this.  Dealing with some of the outdated constructs was on the to-do list - the original draft of much of this was done in Python 2.7 and so it still bears the marks of that.  So the second and third commits are very helpful.

In terms of re-formatting: this is something else that is on my list.  However, although I've been influenced in my own Python style by the philosophy in black, I've decided that I like my coffee black but not my Python code.  There are some places where it pedantically spreads out what could fit on two lines, into a dozen.  Because there is a lot of pending maintenance to be done, my plan is to reformat as I go in dealing with these items, to avoid having the git blame broken solely for formatting.  (I know that the black philosophy is 'just do it all at once and then you never have the formatting problem again'...)

EwoutH:
Succeeded by #294.

PR #287, opened 2022-01-14
CI: Use Python 3.10 instead of 3.1

EwoutH:
The `3.10` is currently interpreted as a number, and thus Python 3.1 is used. Changing it to `'3.10'` as a string fixes this.

### DIFF
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -9,7 +9,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.10]
+        python-version: ['3.10']
 
     steps:
       - uses: actions/checkout@v2
@@ -34,7 +34,7 @@ jobs:
     runs-on: windows-latest
     strategy:
       matrix:
-        python-version: [3.10]
+        python-version: ['3.10']
 
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


EwoutH:
I was wondering why the builds were still passing on Python 3.1, but it was actually using 3.10, because the Python version declared in the matrix wasn't actually used when setting up Python! I fixed that in the second commit. I also added a Python 3.7 run on Ubuntu.

Great work on the Python and Windows CI builds in general though! And glad to see new life blew into Gambit.

tturocy:
Yes, I had noticed that cosmetic problem that was reporting the wrong Python version.

Could you please squash, and also while you're at it add 3.7 to the Windows as well?  I hadn't done so originally because, until things were building on any version, there was little point in having the build fail on multiple versions....

EwoutH:
In general it's best practice to keep CI configurations a bit sparse, at least when they're ran on every commit. Testing every possible combination quickly eats up a lot of resources, especially when the number of variables and/or their ranges grow.

A common solution is only varying one variable at the time from a base scenario. So Ubuntu with Python 3.10 could be the base, and then you vary all OSes with Python 3.10 and all Python versions with Ubuntu. (also Ubuntu runners are cheaper so that helps)

The build matrix isn't that large, so adding 3.7 to the Windows matrix is perfectly doable, but it's something to keep in mind.

By the way, GitHub has a very nice option to squash commits on merging:
Screenshot_736

tturocy:
Yes, I don't want to over-build.  I had been thinking to put more resource into the Windows side, because I don't routinely have access to a Windows system (and even more rarely one which has a dev toolchain set up).  Maybe I'll change to that down the line.

PR #308, opened 2022-06-10
setup.py: Add project URLs for documentation, source code and issues

EwoutH:
A small addition to `setup.py`, in which the `project_urls` property is defined with links to the documentation, source code and issue tracker.

This allows PyPI to display direct links to the documentation, source code and issue tracker.

Screenshot_741

See https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#project-urls


tturocy:
I wanted briefly to acknowledge this - I need to come back to this in a bit when I have the bandwidth to rationalise the branch structure so I'm certain this is merged everywhere it needs to be.

PR #309, opened 2022-06-10
CI: Update used actions to v3

EwoutH:
Update the checkout, setup-python, upload-artifact and download-artifact actions to v3


tturocy:
I noticed that the Windows build is now failing (and also raising new deprecation warnings).  I expect this is nothing to do with moving the actions to v3 and probably to do with the windows-latest having been changed sometime recently?

PR #325, opened 2022-11-02
Wheel CI: Run more often, automatically deploy to PyPI

EwoutH:
Improve the wheel CI by:
- Running more often, not only on tags, but also on pushes, pull requests and once a week, to test the workflow. Artifacts and deployment to PyPI only happens when a tag is created.
- Add a step that automatically deploy wheels to PyPI. It only deploys to PyPI when a tag is created.

To deploy to PyPI is uses the [PyPI publish GitHub Action](https://github.com/pypa/gh-action-pypi-publish). It uses the [API token](https://pypi.org/help/#apitoken) feature of PyPI, which is recommended to restrict the access the action has.

The secret used in `${{ secrets.PYPI_API_TOKEN }}` needs to be created on the settings page. See [Creating & using secrets](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/creating-and-using-encrypted-secrets).

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -1,9 +1,10 @@
 name: pygambit wheels
-
 on:
   push:
-    tags:
-      - 'v*'
+  pull_request:
+  workflow_dispatch:
+  schedule:
+    - cron: '0 6 * * 1'
 
 jobs:
   windows:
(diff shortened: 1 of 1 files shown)


EwoutH:
Hi @tturocy! Would you be able to take a look at this PR? I think it's a nice quality of life feature and could make deploying new releases easier.

tturocy:
Many thanks @EwoutH.  This will definitely be useful - but one of the first jobs on my list with our new funding is going to be looking at our whole packaging-and-dissemination plan.  (It's part of why we didn't yet get around to releasing some of the things on branches.). We're currently recruiting someone who's going to have primary responsibility for that; it just makes sense for us to get them onboard first!

EwoutH:
Glad to hear funding is secured!

tturocy:
Thanks for this.  Separately we've now already adopted some of the changes in other commits.  The only part we haven't done is automating the PyPI upload - we might decide to do that at one point but we're having a think about the whole release automation tooling in the round, so if we do decide to adopt that we'll do it separately at some point down the line.

PR #367, opened 2023-10-02
ENH-355: Use the _resolve_node function of the class Game to belief f…

VarsosC:
…unction of the class MixedBehaviorProfile.


src/pygambit/behav.pxi
```diff
@@ -264,7 +264,7 @@ class MixedBehaviorProfile:
         """
         if node.game != self.game:
             raise MismatchError("belief: node must be part of the same game as the profile")
-        return self._belief(node)
+        return self._belief(self.game._resolve_node(node, 'node'))
```
tturocy, line 267: The second argument to the `_resolve_*` functions is the name of the function being called (for use in formatting any exceptions that are being raised).  So in this case it would be `'belief'` as the name of the function, instead of `node`.

tturocy (changes requested):
Please see comment.

src/pygambit/behav.pxi
VarsosC: I have addressed this comment.

VarsosC (commented):
Changes done

tturocy (changes requested):

PR #368, opened 2023-10-02
ENH-353-initialisation of mixed profiles on creation

VarsosC:


src/pygambit/game.pxi
```diff
@@ -672,16 +672,20 @@ class Game:
                     mspr[s] = Rational(v)
             return mspr
 
-    def mixed_behavior_profile(self, rational=False) -> MixedBehaviorProfile:
-        """Create a behavior strategy profile over the game.
-
-        The profile is initialized to uniform randomization for each player
-        over their actions at each information set.
+    def mixed_behavior_profile(self, data=None, rational=False) -> MixedBehaviorProfile:
+        """Returns a behavior strategy profile `MixedBehaviorProfile` over the game,
```
tturocy, line 676: I think it's useful as a convention to try to keep the first sentence of the docstring concise, as that's what the user will see first in the documentation.  Then after a blank line, we can be more verbose about the operation of the function.  This is the general pattern numpy/scipy/pandas &c follow.

src/pygambit/game.pxi
```diff
@@ -672,16 +672,20 @@ class Game:
...
+        initialized to uniform randomization for each player over his actions at each
```
tturocy, line 677: The docstring repeats itself; it should say what happens when `data` are specified.

src/pygambit/game.pxi
```diff
@@ -672,16 +672,20 @@ class Game:
...
         Parameters
         ----------
-        rational
-            If True, probabilities are represented using rational numbers; otherwise
+	data
+            A nested list (or compatible type) with the same dimension as the action set of the game,
+            specifying the probabilities of the strategies.
+        
+	rational
+            If `True`, probabilities are represented using rational numbers; otherwise
```
tturocy, line 688: It appears that numpydoc convention is that when referring to Python constants like True, False, None and so on, not to enclose them in any back-ticks at all (as in the original version of the docstring)

src/pygambit/game.pxi
```diff
@@ -693,6 +697,20 @@ class Game:
             if not rational:
                 mbpd = MixedBehaviorProfileDouble()
                 mbpd.profile = make_shared[c_MixedBehaviorProfileDouble](self.game)
+                if data is None:
```
tturocy, line 700: A mixed behavior profile is a probability distribution over actions, not nodes.   So the `data` parameter should have the same dimension as the (nested) tuples of (player, infoset, action).  That is, the top level should have length equal to number of players, then for each player, the level should have length equal to that player's number of information sets, and then for each information set, equal to the number of actions.

src/pygambit/tests/test_behav.py
```diff
@@ -7,7 +7,7 @@ def setUp(self):
             "test_games/mixed_behavior_game.efg"
         )
         self.profile_double = self.game.mixed_behavior_profile()
-        self.profile_rational = self.game.mixed_behavior_profile(True)
+        self.profile_rational = self.game.mixed_behavior_profile(None, True)
```
tturocy, line 10: It would be useful to add a test of creating the behavior profile set to a specific distribution, exercising the new code.

tturocy (changes requested):
Please see the individual comments for suggestions/requests.

VarsosC:
Thanks for the suggestions and the requests.

tturocy (changes requested):
Currently this pull request has no changes at all (other than changing whether src/pygambit/game.pxi has a newline at the end of the file, and adds a superfluous `import pygambit` at the start of `test_behav.py`.

src/pygambit/tests/test_behav.py
```diff
@@ -8,14 +8,18 @@ def setUp(self):
         self.game = gbt.Game.read_game(
             "test_games/mixed_behavior_game.efg"
         )
-        self.profile_double = self.game.mixed_behavior_profile(False)
-        self.profile_rational = self.game.mixed_behavior_profile(True)
+        self.profile_double = self.game.mixed_behavior_profile(None, rational=False)
```
tturocy, line 11: The `None` here is unnecessary given that the `rational=` parameter name has been added

src/pygambit/tests/test_behav.py
```diff
@@ -701,3 +705,59 @@ def test_action_value_with_chance_player_action(self):
...
+        assert profile[self.game.players[2]] == data_rational[2]
+
+        data_double = [[[1/5, 4/5], [1/4, 3/4]], [[1, 0]]]
+
+        profile = self.game_with_chance.mixed_behavior_profile(data_double, False)
+
+        assert profile[self.game_with_chance.players[0]] == data_double[0]
+        assert profile[self.game_with_chance.players[1]] == data_double[1]
+
+    def test_probabilities_infoset_by_label_for_specific_distribution(self):
```
tturocy, line 736: This test isn't necessary.

src/pygambit/tests/test_behav.py
```diff
@@ -701,3 +705,59 @@ def test_action_value_with_chance_player_action(self):
...
+        data_double = [[[1/5, 4/5]], [[1/4, 3/4]], [[1, 0]]]
+        profile = self.game.mixed_behavior_profile(data_double, False)
+
+        assert profile["Infoset 1:1"] == data_double[0][0]
+        assert profile["Infoset 2:1"] == data_double[1][0]
+        assert profile["Infoset 3:1"] == data_double[2][0]
+
+    def test_normalize(self):
```
tturocy, line 744: This test isn't necessary

tturocy (changes requested):
I have made a few small comments.  However, because the test suite has been rewritten for `MixedBehaviorProfile` the tests aren't going to merge.  I would suggest making the changes on a new branch from latest `master`. - you should just be able to copy-paste in the changes in `game.pxi` and then write the test of creating with a given distribution in the new style.

PR #369, opened 2023-10-02
ENH-354: add realiz_prob in pygambit/behav.pxi

VarsosC:


src/pygambit/behav.pxi
```diff
@@ -266,7 +266,7 @@ class MixedBehaviorProfile:
         """
         if node.game != self.game:
             raise MismatchError("belief: node must be part of the same game as the profile")
-        return self._belief(node)
+        return self._belief(self.game._resolve_node(node, 'node'))
```
tturocy, line 269: This change has bled over from another PR/issue.

src/pygambit/behav.pxi
```diff
@@ -324,7 +324,36 @@ class MixedBehaviorProfile:
             If `action` is a string and no action in the game has that label.
         """
         return self._action_value(self.game._resolve_action(action, 'action_value'))
-
+	
+    def realiz_prob(self, player: typing.Union[Player, str], action: Action, node: Node) -> float:
```
tturocy, line 328: A realisation probability is a property of a node only.  The function we're calling in C++ is src/games/behav.h:174:

```
  const T &GetRealizProb(const GameNode &node) const;
```

So the argument should just be a `node: typing.Union[Node, str]` (because we want the usual `_resolve_node` mechanism to work for this)

src/pygambit/behav.pxi
```diff
@@ -324,7 +324,36 @@ class MixedBehaviorProfile:
...
+        MismatchError
+            If `player` is a `Player` from a different game.
+            If `action` is an `Action` from a different game.
+            If `node` is not in the same game as the profile.
+        KeyError
+            If `player` is a string and no player in the game has that label.
+            If `action` is a string and no action in the game has that label.
+            If `node` is not in the same game as the profile.
+        """
+        return RealizationProbs(self.profile, self.player, self.game._resolve_action(action, 'action_value'),  self.game._resolve_node(node, 'node'))
```
tturocy, line 355: What is this calling?  There's no function called `RealizationProbs` anywhere.

The best way to write this function is to imitate the `infoset_prob` function.  The only difference between `realiz_prob` and `infoset_prob` is that `realiz_prob` is for one node and `infoset_prob` is the sum over all nodes in an infoset.  But otherwise the structure of the code's going to be identical - it just calls the node version of `GetRealizProb` from the C++ instead of the infoset version.

The only tiny wrinkle is that when updating `gambit.pxd` you'll need to update the wrapping - `GetRealizProb` is currently for Infoset.  Probably the best thing to do is to make `GetRealizProb` refer to Node and rename the `GetRealizProb` for Infoset to `GetInfosetProb`.   If you haven't done this before in Cython, the syntax for renaming a function in a `.pxd` is like

```
double "GetInfosetProb" GetRealizProb(c_GameInfoset)
```

This will make a call to GetInfosetProb in the Cython actually call the C++ function GetRealizProb (i.e. the bit in quotes is an alias).

tturocy (changes requested):
Please see the comments for details.

VarsosC:
I implemented the GetRealizProb to refer to Node and I rename the GetRealizProb for Infoset to GetInfosetProb.

tturocy (changes requested):
This is looking better.  We didn't need to change the C++ function just right now, but we would probably have wanted to do it eventually anyway so there's no harm done in being proactive.

There aren't any tests included that exercise the new function - we don't need a comprehensive test as yet (that will be done as part of a separate issue) but we should at least check one or two calls work.

src/pygambit/behav.pxi
```diff
@@ -376,6 +376,24 @@ class MixedBehaviorProfile:
...
+            node is determined by finding the node with that label, if any.
+
+        Raises
+        ------
+        MismatchError
+            If `node` is an `Node` from a different game.
+        KeyError
+            If `node` is a string and no node in the game has that label.
+        """
+        return self._realiz_prob(self.game._resolve_node(node, 'node_prob'))
```
tturocy, line 395: The name of the function is `realiz_prob` so we should pass `'realiz_prob'` to `._resolve_node`, instead of `'node_prob`'

tturocy (changes requested):
Looks like just one final very minor item - after that please if you would rebase/squash down to one commit!

src/pygambit/behav.pxi
VarsosC: Done

tturocy (approved):

PR #377, opened 2023-10-09
351 bug check for chance information sets in mixed behavior profiles

rahulsavani:
I have implemented the discussed changes and believe the code in the final commit is good to be merged now; happy to rebase if you suggest that, and to a changelog entry, but would need some guidance on the latter..

### DIFF
--- a/src/pygambit/behav.pxi
+++ b/src/pygambit/behav.pxi
@@ -284,8 +284,13 @@ class MixedBehaviorProfile:
             If `player` is a `Player` from a different game.
         KeyError
             If `player` is a string and no player in the game has that label.
+        ValueError
+            If `player` resolves to the chance player
         """
-        return self._payoff(self.game._resolve_player(player, 'payoff'))
+        resolved_player = self.game._resolve_player(player, 'payoff')
+        if resolved_player.is_chance:
+            raise ValueError("payoff() is not defined for the chance player")
+        return self._payoff(resolved_player)
 
     def infoset_value(self, infoset: typing.Union[Infoset, str]):
         """Returns the expected payoff to the player conditional on reaching an information set,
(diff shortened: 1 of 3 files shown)


src/pygambit/tests/test_behav.py
```diff
@@ -724,3 +729,27 @@ def test_infoset_belief(self):
...
+
+    def test_payoff_with_chance_player(self):
+        """Test to ensure that payoff called with the chance player raises a ValueError"""
+        chance_player = self.game_with_chance.players.chance
+        self.assertRaises(ValueError, self.profile_double_with_chance.payoff, chance_player)
+        self.assertRaises(ValueError, self.profile_rational_with_chance.payoff, chance_player)
+
+    def test_infoset_value_with_chance_player_infoset(self):
+        """Test to ensure that infoset_value called with an infoset of the chance player
+        raises a ValueError"""
```
tturocy, line 741: Style for triple-quoted strings that are over more than one line usually put the closing triple quote on its own line.

src/pygambit/behav.pxi
```diff
@@ -284,8 +284,13 @@ class MixedBehaviorProfile:
             If `player` is a `Player` from a different game.
         KeyError
             If `player` is a string and no player in the game has that label.
+        ValueError
+            If `player` resolves to the chance player
         """
-        return self._payoff(self.game._resolve_player(player, 'payoff'))
+        resolved_player = self.game._resolve_player(player, 'payoff')
+        if resolved_player.is_chance:
+            raise ValueError("It is not possible to call payoff for the chance player")
```
tturocy, line 292: I would suggest saying "payoff() is not defined for the chance player".  After all it's possible to call the function (the calling code just did so!).  It's just that it's not a well-defined concept.  Similar comment on the other functions.

src/pygambit/tests/test_behav.py
```diff
@@ -1,5 +1,4 @@
 import unittest
-
```
tturocy, line 2: The space between the imports was intentional.  The general pattern is

```
imports of standard library
<blank line>
imports of third-party modules
<blank line>
imports of modules from this package
```

tturocy (changes requested):
A few really minor things.

Yes please if you could squash and rebase.

For ChangeLog, you can just add a brief line noting this under the 16.1.0a4 changes that are already listed there.

src/pygambit/tests/test_behav.py
rahulsavani: Understood, thanks.

Actually, I did not remove a blank link. I included the unittest import, along with inheriting from unittest.TestCase, which was not done only for test_behav.py and test_exten.py when I created my branch. I see that you also fixed this for test_behav.py and moved test_exten.py to pytest.

rahulsavani: I see in the changelog: "Migrated to pytest for testing of pygambit."  

There's a few that have not been converted: test_behav.py, test_nash.py, test_game_resolve_functions.py, test_file.py.

PR #386, opened 2023-10-17
.belief calls function of Game class

VarsosC:


tturocy (approved):
This is good.  I have squashed the commits so there is just one, and re-wrote briefly a commit message in what's a better style for long-term reference.

PR #395, opened 2023-10-26
372 bug chance action probabilities after deleting chance action

rahulsavani:
This is now a single commit with a ChangeLog entry, and is so ready for review and possible merging

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,5 +1,11 @@
 # Changelog
 
+## [unreleased] - unreleased
+
+### Fixed
+- When an action at a chance node is deleted the probabilities for the remaining actions are 
+  normalized.
+
 ## [16.1.0a4] - 2023-10-13
 
 ### Changed
--- a/src/games/gametree.cc
+++ b/src/games/gametree.cc
@@ -62,6 +62,10 @@ void GameTreeActionRep::DeleteAction()
     member->children[where]->DeleteTree();
     member->children.Remove(where)->Invalidate();
   }
+
+  if (m_infoset->IsChanceInfoset()) {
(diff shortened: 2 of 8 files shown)


src/games/game.h
```diff
@@ -847,6 +847,8 @@ class GameRep : public BaseGameRep {
   //@{
   /// Set the probability distribution of actions at a chance node
   virtual Game SetChanceProbs(const GameInfoset &, const Array<Number> &) = 0;
+  /// Normalize the probability distribution of actions at a chance node
```
tturocy, line 850: Having a separate `NormaliseChanceProbs` function makes sense, but it doesn't need to be a member of `GameRep`.  It just needs to be present in `TreeGameRep` as that's the only representation (at least at the moment) that implements the operation.   That means changes to the other representation classes can be rolled back.

Further, within `TreeGameRep`, this should be a private member function; there's no reason for external code to call this as having the probabilities be summing to one should already be maintained as an invariant.

tturocy (changes requested):
See the one improvement in terms of localising this change to `TreeGameRep` alone.  Otherwise this looks ready to merge - if you can squash and force-push first that'd certainly be helpful!

tturocy (changes requested):
You've accidentally added the .idea config files in the commit.  We do actually need to have a conversation about which ones we should include - because these can control things like static code checkers and formatters using those.  But that's something for another time!

rahulsavani:
> You've accidentally added the .idea config files in the commit. We do actually need to have a conversation about which ones we should include - because these can control things like static code checkers and formatters using those. But that's something for another time!

Yep I did a git add . a one point and apparently didn't clear up fully. That's been dealt with now and squashed.

PR #402, opened 2023-11-09
Fix bug with Game.write for 'gte' format

AbhijeetKrishnan:
Similar to the bug that got fixed in 90bbf7c, another bug was found that affected the `Game.write_game('gte')` method. A basic test has been added that failed prior to the fix, and passes with it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,3 +36,4 @@ missing
 gambit
 .python-version
 dist
+venv/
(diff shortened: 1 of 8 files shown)


tturocy:
Many thanks.  For this we needed to clean up the commit history so it was a single commit; now merged as 3ff06e26.

PR #421, opened 2023-12-15
Changes in append_move() and append_infoset()

VarsosC:
Here, the user can import a list of valid and existing terminal nodes in the append_move() and the append_infoset() functions. 

Further, the tests test_game, test_node, test_infoset, and test_userguide have been modified in order to use/test the new functionality.

### DIFF
--- a/src/pygambit/game.pxi
+++ b/src/pygambit/game.pxi
@@ -21,6 +21,7 @@
 #
 import itertools
 import pathlib
+import threading
 
 import numpy as np
 from deprecated import deprecated
@@ -886,13 +887,13 @@ class Game:
                 raise KeyError(f"{funcname}(): no strategy with label '{strategy}'")
         raise TypeError(f"{funcname}(): {argname} must be Strategy or str, not {strategy.__class__.__name__}")
 
-    def _resolve_node(self, node: typing.Any, funcname: str, argname: str = "node") -> Node:
-        """Resolve an attempt to reference a node of the game.
+    def _resolve_node(self, nodes: typing.Any, funcname: str, argname: str = "node") -> typing.Iterable[Node]:
+        """Resolve an attempt to reference a group of nodes of the game.
 
         Parameters
         ----------
-        node : Any
-            An object to resolve as a reference to a node.
+        nodes : Any
+            A group of nodes to resolve as a reference to a node.
         funcname : str
             The name of the function to raise any exception on behalf of.
         argname : str, default 'node'
@@ -905,23 +906,42 @@ class Game:
         KeyError
             If `node` is a string and no node in the game has that label.
         TypeError
-            If `node` is not a `Node` or a `str`
+            If `node` is not either a `Node` or a `str` or an `Iterable`.
         ValueError
             If `node` is an empty `str` or all spaces
         """
-        if isinstance(node, Node):
-            if node.game != self:
-                raise MismatchError(f"{funcname}(): {argname} must be part of the same game")
-            return node
-        elif isinstance(node, str):
-            if not node.strip():
-                raise ValueError(f"{funcname}(): {argname} cannot be an empty string or all spaces")
-            for n in self.nodes():
-                if n.label == node:
-                    return n
-            raise KeyError(f"{funcname}(): no node with label '{node}'")
-        raise TypeError(f"{funcname}(): {argname} must be Node or str, not {node.__class__.__name__}")
+        ReferenceSet = []
+        if isinstance(nodes, Node):
+           if nodes.game != self:
+              raise MismatchError(f"{funcname}(): {argname} must be part of the same game")
+           else:
+              return nodes
+        elif isinstance(nodes, str):
+           if not nodes.strip():
+              raise ValueError(f"{funcname}(): {argname} cannot be an empty string or all spaces")
+           for n in self.nodes():
+              if n.label == nodes:
+                 return n
+           raise KeyError(f"{funcname}(): no node with label '{nodes}'")
+        elif hasattr(nodes, '__iter__'):
+           for node in nodes: 
+              if isinstance(nodes, Node):
+                 if nodes.game != self:
+                    raise MismatchError(f"{funcname}(): {argname} must be part of the same game")
+                 else:
+                    ReferenceSet.append(node)
+              elif isinstance(nodes, str):
+                  if not nodes.strip():
+                      raise ValueError(f"{funcname}(): {argname} cannot be an empty string or all spaces")
+                  for n in self.nodes():
+                      if n.label == nodes:
+                         ReferenceSet.append(n)
+                  raise KeyError(f"{funcname}(): no node with label '{nodes}'")
+           return ReferenceSet
 
+        if not isinstance(argname, (Node, str)) and not hasattr(argname, '__iter__'):
+           raise TypeError(f"{funcname}(): {argname} must be either Node, or str, or Iterable")
+		
     def _resolve_infoset(self, infoset: typing.Any, funcname: str, argname: str = "infoset") -> Infoset:
         """Resolve an attempt to reference an information set of the game.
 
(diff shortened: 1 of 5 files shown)


src/pygambit/game.pxi
```diff
@@ -1008,17 +1011,22 @@ class Game:
...
-            raise UndefinedOperationError("append_move(): `node` must be a terminal node")
-        if len(actions) == 0:
-            raise UndefinedOperationError("append_move(): `actions` must be a nonempty list")
-        resolved_node.node.deref().AppendMove(resolved_player.player, len(actions))
-        for label, action in zip(actions, resolved_node.infoset.actions):
-            action.label = label
-
-    def append_infoset(self, node: typing.Union[Node, str],
+        L = []
+        for node in nodes:
```
tturocy, line 1015: The signature of the function allows for a `NodeReference` to be passed.  This line assumes it must be an iterable, so it's not correct.

src/pygambit/game.pxi
```diff
@@ -1008,17 +1011,22 @@ class Game:
...
+           resolved_node = cython.cast(Node, self._resolve_node(node, 'append_move'))
+           resolved_player = cython.cast(Player, self._resolve_player(player, 'append_move'))
+           if len(resolved_node.children) > 0:
+              raise UndefinedOperationError("append_move(): `node` must be a terminal node")
+           if len(actions) == 0:
```
tturocy, line 1020: actions only needs to be checked once, not in each iteration of the loop

src/pygambit/game.pxi
```diff
@@ -1008,17 +1011,22 @@ class Game:
...
+              raise UndefinedOperationError("append_move(): `actions` must be a nonempty list")
+           resolved_node.node.deref().AppendMove(resolved_player.player, len(actions))
```
tturocy, line 1022: The specification states that the operation must be atomic.  Because this is in the loop, if, for example, the first node is valid but the second is not, then this would result in a partial change to the game, which is not what is specified.  All checks must complete first.

src/pygambit/game.pxi
```diff
@@ -1008,17 +1011,22 @@ class Game:
...
-        if len(resolved_node.children) > 0:
-            raise UndefinedOperationError("append_move(): `node` must be a terminal node")
-        if len(actions) == 0:
-            raise UndefinedOperationError("append_move(): `actions` must be a nonempty list")
-        resolved_node.node.deref().AppendMove(resolved_player.player, len(actions))
-        for label, action in zip(actions, resolved_node.infoset.actions):
-            action.label = label
-
-    def append_infoset(self, node: typing.Union[Node, str],
+        L = []
```
tturocy, line 1014: Python variables by convention are snake_case.  By extension of our existing convention this should be called something like `resolved_nodes`

src/pygambit/tests/test_game.py
```diff
@@ -74,6 +74,6 @@ def test_game_dereference_invalid():
     game = gbt.Game.new_tree()
     game.add_player("One")
     strategy = game.players[0].strategies[0]
-    game.append_move(game.root, game.players[0], ["a", "b"])
+    game.append_move([game.root], game.players[0], ["a", "b"])
```
tturocy, line 77: The original test is correct - that you had to change this is an indication that your code is wrong, because the spec states that the new implementation is backwards-compatible.

If a test fails, fix the code don't change the test.

This comment applies equally to a number of the other changes both to the tests and the User Guide - in particular everything in the User Guide should run unchanged.

tturocy (changes requested):

src/pygambit/game.pxi
```diff
@@ -21,6 +21,7 @@
 #
 import itertools
 import pathlib
+import threading
```
tturocy, line 24: There is no need to introduce threading (and threading.Lock doesn't do what you seem to think it does anyway).

src/pygambit/game.pxi
```diff
@@ -886,13 +887,13 @@ class Game:
                 raise KeyError(f"{funcname}(): no strategy with label '{strategy}'")
         raise TypeError(f"{funcname}(): {argname} must be Strategy or str, not {strategy.__class__.__name__}")
 
-    def _resolve_node(self, node: typing.Any, funcname: str, argname: str = "node") -> Node:
-        """Resolve an attempt to reference a node of the game.
+    def _resolve_node(self, node: typing.Any, funcname: str, argname: str = "node") -> typing.Iterable[Node]:
```
tturocy, line 890: You cannot change the return type of this function, for two reasons:

(1) Sometimes you are returning a node and sometimes an iterable of nodes.  So your implementation does not match the type hints.

(2) This will break other functions that call this, because at the moment we do not support lists of nodes in other functions.  We may do so eventually but for now it is a bug to allow it.

src/pygambit/game.pxi
```diff
@@ -905,23 +906,42 @@ class Game:
         KeyError
             If `node` is a string and no node in the game has that label.
         TypeError
-            If `node` is not a `Node` or a `str`
+            If `node` is not either a `Node` or a `str` or an `Iterable`.
         ValueError
             If `node` is an empty `str` or all spaces
         """
+        ReferenceSet = []
```
tturocy, line 913: As I mentioned before variable names must *always* be snake_case.  I please don't want to have to ask again!

src/pygambit/game.pxi
```diff
@@ -905,23 +906,42 @@ class Game:
...
         elif isinstance(node, str):
-            if not node.strip():
-                raise ValueError(f"{funcname}(): {argname} cannot be an empty string or all spaces")
-            for n in self.nodes():
-                if n.label == node:
-                    return n
-            raise KeyError(f"{funcname}(): no node with label '{node}'")
-        raise TypeError(f"{funcname}(): {argname} must be Node or str, not {node.__class__.__name__}")
-
+           if not node.strip():
```
tturocy, line 920: This is all duplicated code.  Duplicating nontrivial code is not acceptable.

src/pygambit/game.pxi
```diff
@@ -994,48 +1014,79 @@ class Game:
...
-                    player: typing.Union[Player, str],
-                    actions: typing.List[str]) -> None:
-        """Add a move for `player` at the terminal node `node`.  `node` becomes part of
+    NodeReference = typing.Union[Node, str]
+    NodeReferenceSet = typing.Iterable[NodeReference]
+
+    def append_move(self, nodes: typing.Union[NodeReference, NodeReferenceSet],
+                          player: typing.Union[Player, str],
+                          actions: typing.List[str]) -> None:
+        """Add a move for `player` at terminal`nodes`.  All elements of `nodes` become part of
```
tturocy, line 1023: Typo, missing space before `nodes`.

src/pygambit/game.pxi
```diff
@@ -994,48 +1014,79 @@ class Game:
...
-            raise UndefinedOperationError("append_move(): `node` must be a terminal node")
         if len(actions) == 0:
-            raise UndefinedOperationError("append_move(): `actions` must be a nonempty list")
-        resolved_node.node.deref().AppendMove(resolved_player.player, len(actions))
-        for label, action in zip(actions, resolved_node.infoset.actions):
-            action.label = label
-
-    def append_infoset(self, node: typing.Union[Node, str],
+           raise UndefinedOperationError("append_move(): `actions` must be a nonempty list")
+        lock = threading.Lock()
```
tturocy, line 1037: As noted above this is not needed (and actually has no effect due to the GIL).

src/pygambit/game.pxi
```diff
@@ -994,48 +1014,79 @@ class Game:
...
+        if hasattr(nodes, '__iter__'):
```
tturocy, line 1038: Duplicated code again.

src/pygambit/game.pxi
```diff
@@ -994,48 +1014,79 @@ class Game:
...
+           with lock:
+              L = []
```
tturocy, line 1040: Variables must always be snake_case

src/pygambit/game.pxi
```diff
@@ -994,48 +1014,79 @@ class Game:
...
+              for nd in nodes:
+                 rs = cython.cast(Node, self._resolve_node(nd, 'append_move'))
+                 if rs in L:
+                    raise UndefinedOperationError("append_move(): `nodes` resolve to the same terminal node")
+                 if len(rs.children) > 0:
+                    raise UndefinedOperationError("append_move(): `node` must be a terminal node") 
+                 L.append(rs) 
+              for nd in nodes:
+                 resolved_node = cython.cast(Node, self._resolve_node(nd, 'append_move'))
```
tturocy, line 1049: I do not follow why nodes need to be resolved twice.

src/pygambit/game.pxi
```diff
@@ -994,48 +1014,79 @@ class Game:
...
+            If an element in `nodes` is a `Node` from a different game, or `infoset` is an `Infoset` from a
             different game.
         """
-        resolved_node = cython.cast(Node, self._resolve_node(node, 'append_infoset'))
         resolved_infoset = cython.cast(Infoset, self._resolve_infoset(infoset, 'append_infoset'))
-        if len(resolved_node.children) > 0:
-            raise UndefinedOperationError("append_move(): `node` must be a terminal node")
-        resolved_node.node.deref().AppendMove(resolved_infoset.infoset)
-
+        if hasattr(nodes, '__iter__'):
```
tturocy, line 1074: Same comments as above on duplication of code.  Further, nontrivial code is now being duplicated between `append_move` and `append_infoset` and one of the items in the spec was to ensure this didn't happen.

src/pygambit/tests/test_node.py
```diff
@@ -213,3 +219,66 @@ def test_node_move_across_games(self):
...
+            pygambit.UndefinedOperationError,
+            self.basic_game.append_move,
+            [self.basic_game.root.children[0],
+             self.basic_game.root.children[0]],
+            "Player 3",
+            ["B", "F"]
+        )
+        nodes_posterior = self.basic_game.nodes()
+
+        assert nodes_posterior == nodes_prior
```
tturocy, line 239: This technically doesn't check whether *any* changes have been made to the game.  A convenient way to check this would be to serialise it to the `.efg` format and compare that as a "canonical" representation.

src/pygambit/tests/test_node.py
```diff
@@ -213,3 +219,66 @@ def test_node_move_across_games(self):
...
+        self.basic_game.add_player("Player 4")
+        nodes2 = [self.basic_game.root.children[1],
+                  self.basic_game.root.children[0].children[1]]
+
+        self.assertRaises(pygambit.UndefinedOperationError, self.basic_game.append_move,
+                          nodes2, "Player 4", ["FU", "BD"])
+
+    def test_ghost_addition(self):
+        """Take a list of nodes. The first item of the list is
+        a terminal node, and the second one succeeds the first.
```
tturocy, line 269: This does not make any sense.  By definition a terminal node has no successors.

src/pygambit/tests/test_userguide.py
```diff
@@ -23,6 +23,7 @@ def test_myerson_poker():
     g.append_move(g.root, g.players.chance, ["King", "Queen"])
     for node in g.root.children:
         g.append_move(node, "Alice", ["Raise", "Fold"])
+
```
tturocy, line 26: Please avoid adding spurious whitespace.

tturocy (changes requested):
Please see comments in the diffs.

Issue #387, opened 2023-10-19
ENH: Improve cacheing of computed information in MixedBehaviorProfile


tturocy:
## Overview

The `MixedBehaviorProfile` implementation is somewhat clever, in that it has a mechanism for cacheing values of most of the interesting quantities pertaining to probabilities and expected payoffs at various points in a game tree.

However, it's not very clever in some other ways:
1.  When it does compute these values, it computes and caches all of them - even if they are not needed to compute the requested value.  For example, if a node realisation probability is requested, it goes on to compute information probabilities, beliefs, action values, and so on.  It is really not clear that this is a good idea, especially for large games.
2.  Computing the expected payoff to a player does not use the cacheing mechanism at all, even though doing so computes the same information.  (Further it means that two calls to getting the expected payoff do the calculation over again).

## Desired outcome

1.  Review the computation of cached quantities, and implement a better system which computes only what is needed.  This requires thinking about what depends on what.
2. Think about what it is worth cacheing versus what is not.  For example, node realisation probabilities should probably be cached, because these are a natural by-product of computing expected payoffs - any traversal of the tree will generate them.  However, it might not be worth cacheing information set probabilities or beliefs, as these are quite easy to compute once you've got node realisation probabilities.
3. Related to the above point, at the moment the arrays which store the cached quantities are allocated when the profile is created.  This is irrespective of whether they are ever used.  So, allocating `MixedBehaviorProfile`s is not very lightweight.  A mechanism should be implemented to allocate the required cacheing object only when required.  A possible candidate for this would be to use a `std::pair<bool, std::unique_ptr<type>>` for whatever `type` needs to be used to store the data.  The first element would be `true` only if the cached value is valid; the second one would be allocated only the first time the cache is needed.  Note that once allocated there's no need to deallocate the space; just mark it as "dirty" but re-use the space to store re-computed values later if needed.

## See also

#375 calls for improving the test suite for `MixedBehaviorProfile`s.  It would be natural either to take on this issue after that is completed, as having good tests in place will confirm any changes work correctly.

PR #441, opened 2024-02-01
using std::map for all the caching data structures in MixedBehaviorPr…

rahulsavani:
…ofile


src/games/behavmixed.imp
```diff
@@ -140,44 +109,28 @@ template <class T>
...
-  m_realizProbs = (T) 0.0;
-  m_beliefs = (T) 0.0;
-  m_nodeValues = (T) 0.0;
-  m_infosetValues = (T) 0.0;
-  m_actionValues = (T) 0.0;
-  m_regret = (T) 0.0;
-
+    m_gameversion(p_profile.GetGame()->GetVersion())
+{
   ((Vector<T> &) *this).operator=((T)0);
```
tturocy, line 114: Should replace this line with

```
static_cast<Vector<T> &>(m_probs) = T(0);
```

A very good example of a line where three things can be changed, all of which used to be done because once upon a time compilers didn't like certain expressions...!

src/games/behavmixed.imp
```diff
@@ -140,44 +109,28 @@ template <class T>
...
   GameRep *game = m_support.GetGame();
 
   for (auto player : game->GetPlayers()) {
-    m_nvals = (T) 0;
-    m_bvals = (T) 0;
-
+    // reset to 0 or clear of map_nvals and map_bvals needed
+    //for (auto &p: map_nvals) p.second = (T) 0;
+    //for (auto &p: map_bvals) p.second = (T) 0;
+    map_bvals.clear();
```
tturocy, line 124: Looking at this, `nvals` and `bvals` are scratch data, and so they really shouldn't be members of the class.  Instead it would seem to make more sense to have them be declared and initialised here in this function and then passed to the recursive calls.

They would have been made members of the class simply because in the old way of doing things allocation was expensive and so it seemed to make sense to have them be allocated once as scratch arrays and kept around.

src/games/behavmixed.imp
```diff
@@ -368,8 +321,6 @@ T MixedBehaviorProfile<T>::GetLiapValue(bool p_definedOnly) const
 
   T x, result = ((T) 0), avg, sum;
 
-  // HACK: force it to recompute data.  FIX THIS.
```
tturocy, line 371: Just confirming as an aide-memoire that this comment was referring to the use of this function in `liap_solve`, and that `liap_solve` used to side-step the calls that invalidated these cached data.  So it is completely OK to be rid of this hack now.

src/games/behavmixed.imp
```diff
@@ -368,8 +321,6 @@ T MixedBehaviorProfile<T>::GetLiapValue(bool p_definedOnly) const
...
-  m_cacheValid = false;
   ComputeSolutionData();
 
   for (int i = 1; i <= m_support.GetGame()->NumPlayers(); i++) {
```
tturocy, line 326: You might just want to crib the implementation of this from `liap_solve`.  We have an open question about how to deal with normalisation of the profiles and/or negative probabilities.

src/games/behavmixed.imp
```diff
@@ -416,7 +365,7 @@ T MixedBehaviorProfile<T>::GetInfosetProb(const GameInfoset &iset) const
   ComputeSolutionData();
   T prob = (T) 0;
   for (int i = 1; i <= iset->NumMembers(); i++) {
```
tturocy, line 367: Definitely use `GetMembers()`.

src/games/behavmixed.imp
```diff
@@ -426,23 +375,28 @@ const T &MixedBehaviorProfile<T>::GetBeliefProb(const GameNode &node) const
...
 
 template <class T>
 Vector<T> MixedBehaviorProfile<T>::GetPayoff(const GameNode &node) const
+// TODO: remove use of Vector
 {
   CheckVersion();
   ComputeSolutionData();
-  return m_nodeValues.Row(node->GetNumber());
+  Vector<T> ret(node->GetGame()->NumPlayers());
+  for (int i = 1; i <= node->GetGame()->NumPlayers(); i++) {
```
tturocy, line 388: For this you can use `std::transform` to get rid of the need to index players by integers.

src/games/behavmixed.imp
```diff
@@ -620,14 +567,13 @@ T MixedBehaviorProfile<T>::DiffNodeValue(const GameNode &p_node,
       // We've encountered the action; since we assume perfect recall,
       // we won't encounter it again, and the downtree value must
       // be the same.
-      return m_nodeValues(p_node->GetChild(p_oppAction->GetNumber())->GetNumber(),
-			  p_player->GetNumber());
+      return map_nodeValues[p_node->GetChild(p_oppAction->GetNumber())][p_player];
```
tturocy, line 570: There's now a version of `GetChild()` that just takes the action directly so we can avoid the `->GetNumber()`

src/games/behavmixed.imp
```diff
@@ -643,130 +589,119 @@ T MixedBehaviorProfile<T>::DiffNodeValue(const GameNode &p_node,
 //             MixedBehaviorProfile<T>: Cached profile information
 //========================================================================
 
+
+// compute realization probabilities for nodes and isets.
+template <class T>
+void MixedBehaviorProfile<T>::ComputePass1_realizProbs(const GameNode &node) const
+{
+  if (node->GetParent()) {
```
tturocy, line 597: A conditional expression would be more compact.

src/games/behavmixed.imp
```diff
@@ -643,130 +589,119 @@ T MixedBehaviorProfile<T>::DiffNodeValue(const GameNode &p_node,
...
+    map_realizProbs[node] = map_realizProbs[(node->GetParent())] * GetActionProb(node->GetPriorAction());
+  }
+  else {
+    map_realizProbs[node] = (T) 1;
+  }
+
+  if (node->GetInfoset()) {
```
tturocy, line 604: You don't actually need to check `node->GetInfoset()`; `GetChildren()` would be empty if `GetInfoset()` returns null.

src/games/behavmixed.imp
```diff
@@ -643,130 +589,119 @@ T MixedBehaviorProfile<T>::DiffNodeValue(const GameNode &p_node,
...
     }
   }
 
   GameInfoset iset = node->GetInfoset();
 
   if (iset) {
     T infosetProb = (T) 0;
-    for (int i = 1; i <= iset->NumMembers(); i++) {
-      infosetProb += m_realizProbs[iset->GetMember(i)->GetNumber()];
+    for (auto i: iset->GetMembers()) {
```
tturocy, line 625: I'd suggest `node` or `member` for the loop variable.  However you can also replace this loop with `std::accumulate` to do the sum.

src/games/behavmixed.imp
```diff
@@ -643,130 +589,119 @@ T MixedBehaviorProfile<T>::DiffNodeValue(const GameNode &p_node,
...
-    for (int pl = 1; pl <= m_support.GetGame()->NumPlayers(); pl++) {
-      m_nodeValues(node->GetNumber(), pl) += static_cast<T>(outcome->GetPayoff(pl));
+    for (auto player: m_support.GetGame()->GetPlayers()) {
+      map_nodeValues[node][player] += static_cast<T>(outcome->GetPayoff(player));
     }
   }
 
   GameInfoset iset = node->GetInfoset();
 
   if (iset) {
```
tturocy, line 623: It would make the code more readable just to check if it's a terminal node and return, rather than having a long nested `if` conditional.

src/games/behavmixed.imp
```diff
@@ -643,130 +589,119 @@ T MixedBehaviorProfile<T>::DiffNodeValue(const GameNode &p_node,
...
-	T &cpay = m_actionValues(act->GetInfoset()->GetPlayer()->GetNumber(),
-				 act->GetInfoset()->GetNumber(),
-				 act->GetNumber());
-	if (infosetProb != infosetProb * (T) 0) {
-	  cpay += m_beliefs[node->GetNumber()] * m_nodeValues(childNode->GetNumber(), iset->GetPlayer()->GetNumber());
-	}
-	else {
-	  cpay = (T) 0;
-	}
+        if (infosetProb != infosetProb * (T) 0) {
```
tturocy, line 653: Can use a conditional ternary expression here.  Also can get rid of the multiply-by-zero indirection (as done elsewhere)

ChangeLog
```diff
@@ -29,6 +29,10 @@
 - Improved __repr__ methods in pygambit for game-related classes
 - Further extension of test suite for mixed behavior profiles to cover new indexing and profile
   order consistency for payoff-related calculations
+- Overhaul of caching in MixedBehaviorProfile: std::map is used as the container for cached
```
tturocy, line 32: Don't think we need this much detail - could just write "Changed implementation of caching in implementation of MixedBehaviorProfile (#387).  Especially for items which aren't user-facing brevity's helpful, and especially when we can just link it to the relevant issue for more information.

tturocy (changes requested):
Just a number of minor improvements, to be expected given the magnitude of the change.

ChangeLog
rahulsavani: I generally agree, but actually in this case we didn't even mention maps in the issue. Still I'll be much briefer.

src/games/behavmixed.imp
rahulsavani: Understood on the open question, but separately shouldn't we only have one implementation? I can crib it for now..

rahulsavani: Ah I now see the comment:

// Liapunov function proper - should be replaced with call to profile once
// the penalty is removed from that implementation.

src/games/behavmixed.imp
rahulsavani: I did the following:

```
  /*
  Vector<T> ret(node->GetGame()->NumPlayers());
  for (int i = 1; i <= node->GetGame()->NumPlayers(); i++) {
    ret[i] = map_nodeValues[node][node->GetGame()->GetPlayer(i)];
  }
  return ret;
  */
  Vector<T> ret(node->GetGame()->NumPlayers());
  std::transform (map_nodeValues[node].begin(), map_nodeValues[node].end(), ret.begin(), 
  						[] (std::pair<GamePlayerRep *, T> const &pair) {return pair.second;});
  return ret;
```

What I am not sure about is why the order that transform goes over the players in the map should be the same as the order in `node->GetGame()->NumPlayers()`?

src/games/behavmixed.imp
rahulsavani: Agreed on loop variable.

There's a couple of other places we could use std::accumulate, however, on values of a map like this, I think the code will be less readable -- is a potential speed-up a reason to do this nonetheless? Or is there actually a nice readable implementation with accumulate?

src/games/behavmixed.imp
```diff
@@ -234,7 +187,7 @@ void MixedBehaviorProfile<T>::UndefinedToCentroid()
     GamePlayer player = efg->GetPlayer(pl);
     for (int iset = 1; iset <= player->NumInfosets(); iset++) {
       GameInfoset infoset = player->GetInfoset(iset);
-      if (GetInfosetProb(infoset) > (T) 0) {
+      if (GetInfosetProb(infoset) > T(0)) {
 	continue;
       }
       T total = (T) 0;
```
tturocy, line 193: It'll be useful to try to catch all the remaining C-style casts like this while we're at it.

src/games/behavmixed.imp
```diff
@@ -243,7 +196,7 @@ void MixedBehaviorProfile<T>::UndefinedToCentroid()
       }
       if (total == (T) 0) {
 	for (int act = 1; act <= m_support.NumActions(pl, iset); act++) {
-	  (*this)(pl, iset, act) = (T) 1.0 / (T) m_support.NumActions(pl, iset);
+	  (*this)(pl, iset, act) = T(1.0) / T(m_support.NumActions(pl, iset));
```
tturocy, line 199: Should use `T(1)` instead of `T(1.0)`.  I think `Rational`'s constructor handles this OK but even better to use an int!

src/games/behavmixed.imp
tturocy: Do the iteration over `Game::GetPlayers()` instead of `map_nodeValues`, and that'll ensure the player order is correct by convention.

src/games/behavmixed.imp
```diff
@@ -620,153 +544,132 @@ T MixedBehaviorProfile<T>::DiffNodeValue(const GameNode &p_node,
...
-	  cpay += m_beliefs[node->GetNumber()] * m_nodeValues(childNode->GetNumber(), iset->GetPlayer()->GetNumber());
-	}
-	else {
-	  cpay = (T) 0;
-	}
-      }
-    }
+  	if (!iset->IsChanceInfoset()) {
+		(infosetProb != T(0)) ?
+			map_actionValues[act] += map_beliefs[node] * map_nodeValues[child][iset->GetPlayer()]:
```
tturocy, line 626: The `map_actionValues[act] +=` should be outside the conditional.

src/games/behavmixed.imp
```diff
@@ -620,153 +544,132 @@ T MixedBehaviorProfile<T>::DiffNodeValue(const GameNode &p_node,
...
-    m_realizProbs[node->GetNumber()] = m_realizProbs[node->GetParent()->GetNumber()] * GetActionProb(node->GetPriorAction());
-  }
-  else {
-    m_realizProbs[node->GetNumber()] = (T) 1;
-  }
+void MixedBehaviorProfile<T>::ComputePass3_infosetValues_regret() const
+{
+  // Populate
+  for (auto player: m_support.GetGame()->GetPlayers()) {
+    for (auto infoset : player->GetInfosets()) {
```
tturocy, line 637: There is a `GetInfosets()` defined directly on the game that would save the iteration over players here.

src/games/behavmixed.imp
```diff
@@ -620,153 +544,132 @@ T MixedBehaviorProfile<T>::DiffNodeValue(const GameNode &p_node,
...
+      map_infosetValues[infoset] = T(0);
+      for (auto action : infoset->GetActions()) {
+        map_infosetValues[infoset] += GetActionProb(action) * map_actionValues[action];
+      }
 
-  if (node->GetInfoset()) {
-    for (int i = 1; i <= node->NumChildren(); i++) {
-      ComputeSolutionDataPass1(node->GetChild(i));
+      T brpayoff = map_actionValues[infoset->GetAction(1)];
```
tturocy, line 643: For future-proofing it'd be better to have

`auto actions = infoset->GetActions()`

and then initialise `brpayoff` with `actions.front()` instead of `infoset->GetAction(1)`.  Then we aren't reliant on anything involving conventions for indexing.

tturocy (changes requested):

src/games/behavmixed.imp
```diff
@@ -36,149 +36,99 @@ template <class T>
...
-        RealizationProbs(p_profile, player->GetNumber(), actions, root);
+        map_bvals[root] = p_profile[strategy];
+        RealizationProbs(p_profile, player, actions, root, map_nvals, map_bvals);
       }
     }
-
-    m_nvals[1] = (T) 1;   // set the root nval
-    BehaviorStrat(player->GetNumber(), root);
+    map_nvals[root] = T(1);   // set the root nval
+    auto root = m_support.GetGame()->GetRoot(); // BehaviorStrat takes GameNode not GameTreeNodeRep*
```
tturocy, line 130: The comment is not really necessary now - it only makes sense if you're looking at old versions of the code.

src/games/behavmixed.imp
```diff
@@ -234,16 +184,16 @@ void MixedBehaviorProfile<T>::UndefinedToCentroid()
     GamePlayer player = efg->GetPlayer(pl);
     for (int iset = 1; iset <= player->NumInfosets(); iset++) {
```
tturocy, line 185: This can all be rewritten to use range-notation loops

src/games/behavmixed.imp
```diff
@@ -260,14 +210,14 @@ MixedBehaviorProfile<T> MixedBehaviorProfile<T>::Normalize() const
...
```
tturocy, line 211: Likewise this can also be rewritten to use range-based loops

src/games/behavmixed.imp
```diff
@@ -620,153 +533,135 @@ T MixedBehaviorProfile<T>::DiffNodeValue(const GameNode &p_node,
...
 }
 
 //========================================================================
 //             MixedBehaviorProfile<T>: Cached profile information
 //========================================================================
 
+
+// compute realization probabilities for nodes and isets.
+template <class T>
+void MixedBehaviorProfile<T>::ComputePass1_realizProbs(const GameNode &node) const
```
tturocy, line 561: Would it be a bit more straightforward to pass in the realisation prob as an argument to the function when calling recursively?

src/games/behavmixed.imp
```diff
@@ -620,153 +533,135 @@ T MixedBehaviorProfile<T>::DiffNodeValue(const GameNode &p_node,
...
+  	return;
+
   GameInfoset iset = node->GetInfoset();
 
-  if (iset) {
-    T infosetProb = (T) 0;
-    for (int i = 1; i <= iset->NumMembers(); i++) {
-      infosetProb += m_realizProbs[iset->GetMember(i)->GetNumber()];
-    }
+//  T infosetProb = T(0);
```
tturocy, line 586: Some commented code still to be rid of

tturocy (changes requested):

src/games/behavmixed.imp
rahulsavani: Left for now.

Issue #459, opened 2024-06-12
Check for valid values of path-following parameters in `logit_solve` and related functions


tturocy:
We have recently exposed some of the algorithm parameters for controlling the numerical continuation in `logit_solve`, `logit_solve_lambda`, and `logit_solve_branch`.  At the moment these aren't being checked for being valid values.  The initial step length should be positive, and the acceleration factor should always be at least one.

ankur-kalita:
heyy @tturocy can you assign this issue to me ?... i am working on it

PR #460, opened 2024-06-19
Validate `first_step` and `max_accel` parameters values in `login_solve` and related functions

lusimeon:
see #459

### DIFF
--- a/src/pygambit/nash.py
+++ b/src/pygambit/nash.py
@@ -573,6 +573,10 @@ def logit_solve(
     """
     if maxregret <= 0.0:
         raise ValueError("logit_solve(): maxregret argument must be positive")
+    if first_step <= 0.0:
+        raise ValueError("logit_solve(): first_step argument must be positive")
+    if max_accel < 1.0:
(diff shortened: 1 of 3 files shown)


tturocy:
Thank you for this.  Could you also please consider adding a few tests to check these, and also to check the `maxregret` parameter?  We acknowledge that the test suite does not yet include confirming that errors are thrown on bad arguments, but these issues are part of the process of doing that.

Also, while you are at it, the function names quoted in `logit_solve_branch` and `logit_solve_lambda` aren't correct, so if you could clean those up as well this would be super useful.

lusimeon:
Hi, thank you for your feedback. I've added tests to checks parameters, and correct the mispelled functions names. Seems good for you?

tturocy:
Thank you - apologies that it has taken so long to come back around to this.

I see only one minor item, which is that the inequality on the test for `max_accel` is inconsistent - sometimes it's weak, sometimes it's strict.  A value of 1.0 is valid for the parameter, so the strict inequality `< 1` would be the correct one.

tturocy:
Merged into `master` at ad47d5443 with many thanks.

PR #462, opened 2024-07-15
Add to arrays function

d-kad:
- Add to_arrays method to Game class that takes as its input an array-like object of array-like objects produces a Gambit game object with a strategic form representation.
- Add tests for the above method
- Docstring for the new function is added


src/pygambit/game.pxi
```diff
@@ -336,6 +336,39 @@ class Game:
         g.title = title
         return g
 
+    def to_arrays(self) -> typing.List[np.array]:
+        """Reverse function to from_arrays:
+
+        To_arrays, for a given game, generates players’ payoff tables represented as numpy arrays.
```
tturocy, line 342: Don't need to repeat the function name, or to say it's for a given game, those are obvious from the definition of the function.  Can just say "Generates the payoff tables for players represented as numpy arrays."

src/pygambit/game.pxi
```diff
@@ -336,6 +336,39 @@ class Game:
         g.title = title
         return g
 
+    def to_arrays(self) -> typing.List[np.array]:
```
tturocy, line 339: Is it worth allowing a `dtype` argument, so the output can be in the user's desired datatype directly?  The current implementation always produces `pygambit.Rational`.  Although we're always going to be quite strict about our representation of numbers, a common use-case for converting to and from `numpy` is to use floating-point numbers.  We could take the view that once someone exports data from `pygambit`, it's on them if they encounter problems with rounding errors!

One possibility would be to support `dtype` as an argument with `float` or `fractions.Fraction` as valid types (noting that `pygambit.Rational` is actually derived from `fractions.Fraction`).  We could also support directly the `numpy` floating-point types.

src/pygambit/game.pxi
```diff
@@ -336,6 +336,39 @@ class Game:
...
+        """Reverse function to from_arrays:
+
+        To_arrays, for a given game, generates players’ payoff tables represented as numpy arrays.
+        The number of produced arrays is equal to the number of players.
+
+        Returns
+        -------
+        list of np.arrays
```
tturocy, line 347: The syntax would be `list of np.array` (the plural is implicit in numpydoc style)

tturocy (changes requested):

src/pygambit/game.pxi
d-kad: I agree, enriching the functionality is user-friendly.

If allowing a dtype argument, how do you see it working?
If dtype=float, we could convert every payoff to float and round (e.g., 2 or 3 decimals).
If dtype=fractions.Fraction, then we could convert every payoff to fraction as well.

Is dtype meant to be a simple string or something more complex?

tturocy: `fractions.Fraction` is the default (or should be), because that's the format we use internally for game data.  If floats were requested, we'd use `float(payoff)`.

I would suggest that for dtype we pass the type itself, that is, `dtype=float` or `dtype=fractions.Fraction`.  Then we could write e.g. `dtype(payoff)` - so basically if `dtype` is a type which can convert a `Fraction` to itself, then it will work, otherwise it would generate an error.

d-kad: Adjusted the code accordingly and added two tests (for float and fraction). Please let me know what you think

src/pygambit/game.pxi
```diff
@@ -336,6 +337,34 @@ class Game:
...
+        dtype : type
+            The desired datatype of the payoff values
+
+        Returns
+        -------
+        list of np.array
+
+        See Also
+        --------
+        from_arrays : Create game from list-like of array-like
```
tturocy, line 353: Don't forget also to add the converse cross-reference to `from_arrays`

src/pygambit/game.pxi
```diff
@@ -336,6 +337,34 @@ class Game:
         g.title = title
         return g
 
+    def to_arrays(self, dtype=fractions.Fraction) -> typing.List[np.array]:
+        """Generate the payoff tables for players represented as numpy arrays.
+        Parameters
+        ----------
+        dtype : type
```
tturocy, line 344: The explanation of this could be a bit more explicit - that it's the type to which payoff values will be converted and the resulting arrays will be of that `dtype`.

Also, we are not testing this parameter at all in the tests yet - either testing for correctness when it should work, and also then testing that it does raise an exception on a type that doesn't work.

tturocy (changes requested):

src/pygambit/game.pxi
d-kad: I added one new test (float) and updated the old one (fraction)

src/pygambit/game.pxi
```diff
@@ -336,6 +337,35 @@ class Game:
...
+        arrays = []
+
+        shape = tuple(len(player.strategies) for player in self.players)
+        for player in self.players:
+            array = np.zeros(shape=shape, dtype=object)
+            for profile in itertools.product(*(range(s) for s in shape)):
+                try:
+                    array[profile] = dtype(self[profile][player])
+                except ValueError:
+                    print("Unsupported type of payoff")
```
tturocy, line 365: The behaviour of this is that it actually suppresses the error.  Yes it prints something, but then the code continues to run and returns arrays, but those arrays will be wrong.

The function needs to raise an exception, but our goal here is to make it raise an exception that will be more useful to users to understand the problem.  I would suggest:

```
raise ValueError(f"Payoff '{self[profile][player]}' cannot be converted to requested type '{dtype}'") from None
```

With this, instead the sequence in the above comment behaves like this:

```
In [3]: g.to_arrays(dict)
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[3], line 1
----> 1 g.to_arrays(dict)

File ~/git/gambit/src/pygambit/game.pxi:365, in pygambit.gambit.Game.to_arrays()
    363             array[profile] = dtype(self[profile][player])
    364         except (ValueError, TypeError):
--> 365             raise ValueError(f"Payoff '{self[profile][player]}' cannot be converted to requested type '{dtype}'") from None
    366     arrays.append(array)
    367 return arrays

ValueError: Payoff '1' cannot be converted to requested type '<class 'dict'>'
```

Which means that we're always raising a `ValueError` (irrespective of what the dtype raises), and then giving a useful error message to the user to understand the problem.

The effect of the `from None` is that it suppresses the original exception; if we don't include it the user gets two exception stack traces (one from the original exception and one from ours).  In most cases, the original stack trace isn't going to be interesting, because the point is that the requested type can't represent the payoff.

tests/test_game.py
```diff
@@ -13,6 +13,65 @@ def test_from_arrays():
     assert len(game.players[1].strategies) == 2
 
 
+def test_empty_array_to_arrays():
```
tturocy, line 16: We still need tests for correctly converting to usual types (`float`, `Rational`), but also a test to ensure raising exception on a bad type (can use the `dict` example from above).

tturocy (changes requested):

src/pygambit/game.pxi
```diff
@@ -336,6 +337,39 @@ class Game:
...
+
+        shape = tuple(len(player.strategies) for player in self.players)
+        for player in self.players:
+            array = np.zeros(shape=shape, dtype=object)
+            for profile in itertools.product(*(range(s) for s in shape)):
+                try:
+                    array[profile] = dtype(self[profile][player])
+                except (ValueError, TypeError, IndexError, KeyError):
+                    raise ValueError(
+                        f"Payoff '{self[profile][player]}' cannot be \
```
tturocy, line 367: Splitting strings like this doesn't quite do what you're probably wanting:

```
File ~/git/gambit/src/pygambit/game.pxi:366, in pygambit.gambit.Game.to_arrays()
    364     array[profile] = dtype(self[profile][player])
    365 except (ValueError, TypeError, IndexError, KeyError):
--> 366     raise ValueError(
    367         f"Payoff '{self[profile][player]}' cannot be \
    368         converted to requested type '{dtype}'"

ValueError: Payoff '1' cannot be \
                        converted to requested type '<class 'dict'>'
```

You should (probably) never have a line continuation backslash inside a string.  In Python you can simply close the string and start one on the next line, and they are automatically catenated.   Further, in an exception message, you probably don't want to enforce a line break in a particular place, because where a line break should happen (if at all) would depend on the width of the user's terminal.

doc/pygambit.user.rst
```diff
@@ -183,6 +183,9 @@ the top level index is the choice of the first player, the second level index of
 and so on.  Therefore, to create a two-player symmetric game, as in this example, the payoff matrix
 for the second player is transposed before passing to :py:meth:`.Game.from_arrays`.
 
+There is a reverse function :py:meth:`.Game.to_arrays` that produces
```
tturocy, line 186: Especially in the user guide, we try to make good use of the fact that we can inline Python code and the documentation generator will run the code and include the results.

It would be good then to illustrate how `to_arrays` works by re-generating the arrays from the previous example game.  Further it would be good to show how to use the `dtype` argument by an example.

tturocy (changes requested):
Just a few minor polish items, quite close now.

doc/pygambit.user.rst
```diff
@@ -183,6 +183,22 @@ the top level index is the choice of the first player, the second level index of
...
+the players' payoff tables given a strategic game. The output is the list of numpy arrays,
+where the number of produced arrays is equal to the number of players.
+
+.. ipython:: python
+
+   m, m_transposed = g.to_arrays()
+   m
+
+There is a parameter `dtype`` passed to :py:meth:`.Game.to_arrays` that configures the data type of
+payoffs in the generated arrays. Supported types are :py:class:`.Rational` (default) and float.
```
tturocy, line 196: Actually this isn't quite true - we support any type which can convert from a Python `Fraction`.  Specifically for example, `str` works just fine!  And in fact that could be a nice example to add as well in addition to the `float`.

tturocy (changes requested):
Just one small suggestion in the docs, otherwise looking good.

The last remaining thing to do is to squash the commits down to one commit with a good commit message - see e.g. https://cbea.ms/git-commit/ for some nice habits on this.

tturocy (approved):

tturocy:
The commit history of this has gotten garbled enough that I have manually re-done it and committed as cc7c97e7.

Issue #458, opened 2024-06-12
Accept rational-precision mixed profiles for `logit_estimate`


tturocy:
At the moment `logit_estimate` expects the strategy profile passed to be of float (double) precision.  There's no conceptual reason these couldn't be of rational precision as well; indeed that may be most natural when working with count data.

Extend this method to allow rational precision profiles, by internally creating a copy of the profile in float precision and passing it on to the implementing function(s).

Manan21st:
I had raised a pr for this issue last week. Please look into it. :)

PR #464, opened 2024-08-04
fix-issue: logit_estimate

Manan21st:
This update allows the logit_estimate function to handle rational numbers (like fractions) by converting them to floats internally. This means you can now use profiles with rational precision without any issues, making it easier to work with count data directly.

this fixes issue- #458

### DIFF
--- a/src/pygambit/qre.py
+++ b/src/pygambit/qre.py
@@ -305,6 +305,20 @@ def logit_estimate(
         as a structural model for estimation: The missing manual.
         SSRN working paper 4425515.
     """
+    def convert_to_float_precision(profile):
+        """Convert a rational precision profile to float precision."""
+        if isinstance(profile, libgbt.MixedStrategyProfile):
+            return libgbt.MixedStrategyProfile(
+                {strategy: float(prob) for strategy, prob in profile.items()}
+            )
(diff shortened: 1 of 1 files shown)


tturocy:
Thank you for your willingness to have a go at this.

The main feedback here is about some principles of software engineering and design.  This is not the way one would go about this.  After all, although we have a specific use case here for being able to convert between dtypes of profiles, this is a generally useful operation to have that can be used elsewhere as well.  So it does not make sense to be adding a special function solely in this one place.

Similarly, from a refactoring perspective, the fact that you are doing `if` conditionals checking types, suggests a "code smell" - it's not something one ordinarily should need to do in an object-oriented setting.  That's another hint that this probably isn't the way to attack the problem.

Instead, a good way to proceed is by adding an appropriate member function to the `Mixed*Profile*` classes which allows you to convert to a different dtype.  A good place to look to here for guidance is `numpy`, and operations like `ndarray.astype`.  `pandas` has a similar operation with `Series`/`DataFrame`.  Basically a mixed behaviour profile or mixed strategy profile is very similar to a `pandas` series with a hierarchical index, so it makes a lot of sense to try to imitate that style.

Once you have a function like that, then you should also remember always to provide appropriate tests.  Making this operation a different function likewise facilitates testing it independently.

PR #476, opened 2025-01-13
Add static wxwidgets to mac installation in github actions

EdwinB12:
This PR adds WXWidgets as a static library for the MacOS build with github actions. 

Three files are changed: 

- tools.yml : Following instructions on wxwidgets webpage, a static version of the wxwidgets is built. 
- configure.ac : Setting an IS_MAC variable ; similar to IS_WIN32
- Makefile.am : Some logic to perform to handle libraries differently depending on the OS. There may be a better way of writing this logic. . 

I've manually tested this by opening the GUI on Macos 13 (and Ted tried on Macos14). GUI opens without a previous wxwidgets installed via brew. I don't have familarity with the software to test further but it appears to work as desired. I see the same behaviour on windows. I haven't tested to linux installation. 

Hope this is helpful! :D

### DIFF
--- a/.github/workflows/tools.yml
+++ b/.github/workflows/tools.yml
@@ -40,7 +40,16 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - name: Install dependencies
-        run: brew install automake autoconf wxwidgets
+        run: |
+            brew install automake autoconf 
+            curl -L -O https://github.com/wxWidgets/wxWidgets/releases/download/v3.2.6/wxWidgets-3.2.6.tar.bz2
+            tar xjf wxWidgets-3.2.6.tar.bz2
+            cd wxWidgets-3.2.6
+            mkdir build-release
+            cd build-release
+            ../configure --disable-shared --disable-sys-libs
+            make -j4
+            sudo make install
       - run: aclocal
       - run: automake --add-missing
       - run: autoconf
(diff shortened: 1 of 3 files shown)


tturocy:
Thanks for looking into this!

This actually is the solution we originally considered and rejected.  The problem is this builds wxWidgets every time the action is triggered, and that's unreasonably expensive.

We had been thinking about considering a way to use artefacts.  But now looking at this again fresh - isn't maybe the best solution just to trigger the static build on release?  We already have this done for building Python wheels - see wheels.yml.  Is that the sensible way to do this as well?  That seems a lot less hacky and fragile than relying on an artefact being built somewhere.

We shouldn't need to change configure.ac and Makefile.am, because the configure script already has a `--with-wxdir=` option that points to an uninstalled build of wxWidgets.  So I think it would just be enough to do `./configure --with-wxdir=wxWidgets-3.2.6/build-release` (or similar) when configuring Gambit in the action.

EdwinB12:
I agree, triggering the static build on release only would be a clean solution. 
I did look into cacheing artefacts briefly but didn't love the idea

tturocy:
I followed up with that idea and implemented it in 624073248, so now the new bespoke static build of wxWidgets only runs on release or cron.  Many thanks for your help @EdwinB12 !

EdwinB12:
Sorry I didn't get this over the line, thanks for picking it up!

tturocy:
No problem - getting the static build of wxWidgets part working was a really useful step that probably would have taken me several slow rounds of debugging to get right.

PR #478, opened 2025-01-24
Change representation of MixedBehaviorProfile and MixedStrategyProfile

d-kad:
Reimplementing `__repr__` method for classes `MixedBehaviorProfile` and `MixedStrategyProfile` in the style of the hierarchical index of `pandas.Series`

### DIFF
--- a/src/pygambit/behavmixed.pxi
+++ b/src/pygambit/behavmixed.pxi
@@ -19,6 +19,7 @@
 # along with this program; if not, write to the Free Software
 # Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA 02111-1307, USA.
 #
+from pygambit.util import MultiIndexSeriesFormatter
 import cython
 from cython.operator cimport dereference as deref
 
@@ -372,7 +373,7 @@ class MixedBehaviorProfile:
         raise ValueError("Cannot create a MixedBehaviorProfile outside a Game.")
 
     def __repr__(self) -> str:
-        return str([self[player] for player in self.game.players])
+        return MultiIndexSeriesFormatter().format(self.as_dict())
 
     def _repr_latex_(self) -> str:
         return (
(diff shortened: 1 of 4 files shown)


tturocy (changes requested):
This is something like what I had in mind, but I had two observations.  The first is probably something we want to deal with straightaway, the second might be suitable for a follow-on issue.

The more immediate thing to look at is how this interacts with `NashComputationReslut`.  For example, with the game `g` from `contrib/games/e02.efg`:

```
In[4]: gbt.nash.enummixed_solve(g)
Out[4]: 
NashComputationResult(method='enummixed', rational=True, use_strategic=True, equilibria=[Player 1 1* 1
         21 0
         22 0
Player 2 1  1
         2  0, Player 1 1* 1
         21 0
         22 0
Player 2 1  1/2
         2  1/2], parameters={})
```

Which is kind of ugly.  We should probably change that to say something like
`equilibria=[<list of N profiles>]`?  @rahulsavani suggestions for the terminology here?

The second thought is whether we want to have any column labels and/or provide other details.  The command-line tool `gambit-lcp` offers "detailed" output for the profiles it finds.  Again working with `e02.efg` here's what it generates for the strategic form:

```
Strategy profile for player 1:
Strategy   Prob          Value
--------   -----------   -----------
      1*             1             1
      21             0             0
      22             0             0
Strategy profile for player 2:
Strategy   Prob          Value
--------   -----------   -----------
       1             1             1
       2             0             1
```

And here's what it does for the extensive form (mixed behaviour profile):

```
Behavior profile for player 1:
Infoset    Action     Prob          Value
-------    -------    -----------   -----------
  (1,1)          R             1             1
  (1,1)          L             0           1/2
  (1,2)          r           1/2             0
  (1,2)          l           1/2             0

Infoset    Node       Belief        Prob
-------    -------    -----------   -----------
  (1,1)          1             1             1
  (1,2)          5             0             0

Behavior profile for player 2:
Infoset    Action     Prob          Value
-------    -------    -----------   -----------
  (2,1)          R           1/2             0
  (2,1)          L           1/2             0

Infoset    Node       Belief        Prob
-------    -------    -----------   -----------
  (2,1)          3             0             0
```

Probably showing beliefs ought to be separate (as we are not yet writing anything that produces a behaviour assessment!) - but showing payoffs as well might be nice.

To be clear in the above I wouldn't recommend the very verbose headers as done in the command-line output - I like having the players done a la pandas' hierarchical indexing as it is much more compact.

Issue #440, opened 2024-02-01
MAINT: Refactor implementation of generation of reduced normal form strategies


tturocy:
## Overview

The `TreeGameRep` class for extensive forms has built-in generation of the set of reduced normal form strategies.  The implementation of this is some of the oldest code in Gambit - variable names can be traced to the implementation circa 1990.  The way this is done is tightly coupled into the data structure, and should be refactored.

The programming task proposed is straightforward, but it would be useful for a person taking this on to have a good understanding of the relationship between an extensive form and its reduced strategic form.

## Details

The members `flag` and `whichbranch` of `GameTreeInfosetRep` and `whichbranch` and `ptr` of `GameTreeNodeRep` exist solely as scratch fields for the computation of the set of reduced normal form strategies.  which are used by `GamePlayerRep::MakeReducedStrats`.  These should be replaced by working data in the form of `std::map` which map from the infoset or node (as appropriate) to the scratch data.

emiliomarquez:
Is this still open or has this been implemented already? If its still open and not been implemented, I am going to try refactoring it

tturocy:
At the moment it's not being actively worked on, so feel free to have a go!

tturocy:
To make this more concrete, we will write a function like this (type names provisional/indicative at this point)

```
using StrategyMap = std::map<GameInfoset, GameAction>;

std::map<GamePlayer, std::list<StrategyMap>> MakeReducedStrategies(const Game &);
```

This is a global-scope function (in the `Gambit` namespace).  It shouldn't need to access the internals of games at all - it should be able to do the calculation entirely using public members of `Game`.

tturocy:
We've settled on a concrete plan for accomplishing this:

* [ ] Add a function in C++ which returns the action at an information set for a given strategy
* [ ] Expose that function via Cython in pygambit
* [ ] Write a pure-Python implementation of the algorithm to compute reduced strategy sets
* [ ] Write tests which confirm the new implementation produces the same results (and therefore we have a test suite)
* [ ] Re-implement the pure-Python algorithm in C++ and replace the old implementation, using the test suite to confirm there are no regressions

tturocy:
The first commit on `issue_440` deals with the first two items above with a temporary interface to the action mapping.  This is enough to enable progress on the broader issue, but will need to be polished before we merge this work into master.

PR #481, opened 2025-03-04
Issue 440

d-kad:
This is work in progress.

As the first step of implementing [issue 440](https://github.com/gambitproject/gambit/issues/440), we compute images in outcomes of the vertices.


tturocy:
This has been implemented in PR 521 (issue #517), where we have termed this concept the "plays" associated with those objects.

Issue #489, opened 2025-03-24
Change `GamePlayerRep::m_infosets` from `Gambit::Array` to `std::vector`


tturocy:
We currently represent the information sets belonging to a player as a `Gambit::Array`.  We have identified that when building large games, this creates a noticeable performance problem, because `Gambit::Array` is not clever about reallocation on `.push_back()`.

The task is to replace `Gambit::Array` with `std::vector`.  The one subtle point is that `Gambit::Array` indexes start at 1 by default, while `std::vector` starts at 0.

To accomplish this migration efficiently, the following is suggested.  First change `m_infosets` to a `std::list`.  `std::list` does not support indexing, so calls to `operator[]` will result in a compile error.  Change any relevant calls to an iterator-based approach, via e.g. `std::next(m_infosets.begin(), index-1)`.  This should allow the code to compile and pass the test suite.  Then, the `std::list` can be changed to `std::vector`; the code should then still compile without further changes.  Finally the `std::next` call(s) can be back-replaced with `m_infosets[index-1]`.

PR #493, opened 2025-03-28
Change type of GamePlayerRep::m_infosets from Array to std::vector

d-kad:
Change type of `GamePlayerRep::m_infosets` from `Array` to `std::vector`, which solves issue #489


tturocy:
Rebased and merged as a65784d0

Issue #499, opened 2025-04-02
BUG: `copy_tree` and `move_tree` reverse source and destination nodes


tturocy:
# Overview

In the implementation of `copy_tree` and `move_tree`, the call to the underlying C++ function reverses the roles of the source and destination nodes.

# Actions

This has been a bug since 16.2, so the fix for this should branch off of `maint16_2` for eventual merging back into that branch.

* Add appropriate tests to the test suite that catch this bug (they are missing!)
* Fix the bug itself (straightforward correction in the final lines of `copy_tree` and `move_tree`, respectively)
* Expand documentation of `copy_tree` in both the C++ and Cython to note the following:
  * It is valid for `dest` to be a descendant of `src`; in such a case the copying terminates when the original terminal node is reached.
  * Copied nodes are kept in the same information set as the originals.

PR #502, opened 2025-04-04
Issue 499

d-kad:
Solves #499 by fixing the right order of source and destination nodes in copy_tree and move_tree methods. 
Adds a test comparing subtrees after successful copy_tree.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,5 +1,11 @@
 # Changelog
 
+## [16.2.2] - unreleased
+
+### Fixed
+- `Game.copy_tree` and `Game.move_tree` implementations reversed the roles of the
+  `src` and `dest` nodes (#499)
+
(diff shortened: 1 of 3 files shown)


tturocy:
I'm not following the intent of `DeleteOutcome`.  The source node shouldn't be changed..  But further, `DeleteOutcome` does not just un-set the outcome at the node, it completely removes the outcome from the game - which we definitely don't want!

For this because it's a bug that's existed since 16.2, we want to base this on and merge this into `maint16_2` - from there it will get merged forward into newer versions.

Finally, don't forget to add a `ChangeLog` entry (as this is a user-visible change) and mentioning the issue number.

tturocy:
Merged at:
* dab6a39 in maint16_2
* 0bb6ddf in maint16_3
* e487cb4 in master

Issue #491, opened 2025-03-25
Extend `GameTreeRep` to keep track of the total number of nodes in the game.


tturocy:
## Overview

As part of #440, it will be useful to keep track of how many terminal nodes there are in a game tree.  Alongside that, we can observe that while `GameRep` has a function to return the number of nodes in a game, this actually works via a recursive function call that traverses the tree whenever called, which is.... not great.

## Plan

* Add a member `m_numNodes` to `GameTreeRep`.  This should always have the current number of nodes.  This should be referred to by `NumNodes()`.
* Replace the existing implementation of `Game.nodes` in `game.pxi` to be a collection similar to `GamePlayers`, `GameOutcomes` and so on.  `GameNodes.__len__()` should call the new `NumNodes` function.  Iteration over nodes can be dealt with for now via a recursive generator (using e.g. `yield from`).
* Add appropriate tests for this to the `pygambit` test suite - checking all operations that change the number of nodes in the tree.

d-kad:
the parts of the plan dealing with non-terminal nodes are delegated to #503

PR #504, opened 2025-04-11
Issue 491

d-kad:
- add `GameTreeRep::m_numNodes` field and expose it in `NumNodes`
- reimplement `Game.nodes` as a property of class `Game`
- add tests for new functionality of `GameNodes.__len__`

solves #491

### DIFF
--- a/src/pygambit/game.pxi
+++ b/src/pygambit/game.pxi
@@ -161,6 +161,40 @@ def read_agg(filepath_or_buffer: typing.Union[str, pathlib.Path, io.IOBase]) -> 
 
 
 @cython.cclass
+class GameNodes:
+    """Represents the set of nodes in a game."""
+    game = cython.declare(c_Game)
+
+    def __init__(self, *args, **kwargs) -> None:
+        raise ValueError("Cannot create GameNodes outside a Game.")
+
+    @staticmethod
+    @cython.cfunc
+    def wrap(game: c_Game) -> GameNodes:
+        obj: GameNodes = GameNodes.__new__(GameNodes)
+        obj.game = game
+        return obj
+
+    def __repr__(self) -> str:
+        return f"GameNodes(game={Game.wrap(self.game)})"
+
+    def __len__(self) -> int:
+        """The number of nodes in the game."""
+        if not self.game.deref().IsTree():
+            return 0
+        return self.game.deref().NumNodes()
+
+    def __iter__(self) -> typing.Iterator[Node]:
+        def dfs(node):
+            for child in node.children:
+                yield from dfs(child)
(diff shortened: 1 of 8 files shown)


d-kad:
In `test_realiz_prob_nodes_reference` (`test_behav.py`) I had to change associations between node indices and probabilities as the traversal order for nodes has changed, seemingly due to the added `dfs` in `GameNodes.__iter__`

@tturocy should we keep it as is or should the old order be restored?

src/pygambit/game.pxi
```diff
@@ -667,6 +701,11 @@ class Game:
         """The set of outcomes in the game."""
         return GameOutcomes.wrap(self.game)
 
+    @property
+    def nodes(self) -> GameNodes:
+        """The set of nodes in the game."""
+        return GameNodes.wrap(self.game)
```
tturocy, line 707: I would suggest a few things here on the documentation:

* Add a `..versionchanged` note to indicate that from 16.4 this is a property and not a function, because this is a compatibility-breaking change
* Keep a version of the sentence about iteration in the docstring, confirming the order in which nodes are iterated (this isn't changed)

tests/test_extensive.py
```diff
@@ -57,7 +57,7 @@ def test_game_add_players_nolabel():
 
 def test_game_num_nodes():
     game = games.read_from_file("basic_extensive_game.efg")
-    assert len(game.nodes()) == 15
+    assert len(game.nodes) == 15
```
tturocy, line 60: Because we are now cacheing the count of nodes, this test alone isn't adequate - we really should be checking node counts after any operation that adds or deletes nodes.

tturocy (changes requested):

tturocy:
> In `test_realiz_prob_nodes_reference` (`test_behav.py`) I had to change associations between node indices and probabilities as the traversal order for nodes has changed, seemingly due to the added `dfs` in `GameNodes.__iter__`
> 
> @tturocy should we keep it as is or should the old order be restored?



> In `test_realiz_prob_nodes_reference` (`test_behav.py`) I had to change associations between node indices and probabilities as the traversal order for nodes has changed, seemingly due to the added `dfs` in `GameNodes.__iter__`
> 
> @tturocy should we keep it as is or should the old order be restored?

This is happening because of the order you're yielding nodes in the recursive iterator.  Switch it so that you yield the node first then its children - this will match the previous behaviour.  No need to change behaviour here unnecessarily!

This is a nice example though of how that particular test is a bit fragile and why having something like node paths will be an important improvement to develop.

tturocy:
Merged as 0a04bab.

PR #510, opened 2025-04-22
change m_players in GameRep from Array to std::vector

### LINKED ISSUE #496: Change `GameXXXRep::m_players` from `Gambit::Array` to `std::vector` and move to `GameRep::m_players`
This has two steps - which are sufficiently closely related they warrant being dealt with at the same time.

The *second* step is closely analogous to #488 and #489, involving replacing a (1-based) `Gambit::Array` with a (0-based) `std::vector`.  See those issues for a suggested approach to ensuring this works correctly.

The *first* step is to remove some code duplication.  `m_players` actually appears only in classes derived from `GameRep`, namely `GameExplicitRep`, `GameAGGRep`, and `GameBAGGRep`.  Inspection of those code however shows that actually these are completely redundant - they could just as easily be defined in `GameRep` and avoid code duplication.  As a first step, complete this change, representing the `Gambit::Array` of players in `GameRep` and removing the duplicated code.  In particular, this means that `NumPlayers` and `GetPlayer` no longer need to be virtual.

d-kad:
Closes #496

### DIFF
--- a/src/games/game.cc
+++ b/src/games/game.cc
@@ -200,6 +200,8 @@ size_t GamePlayerRep::NumSequences() const
 //                            class GameRep
 //========================================================================
 
+GamePlayer GameRep::GetPlayer(int pl) const { return m_players.at(pl - 1); }
+
 Array<GamePlayer> GameRep::GetPlayers() const
 {
   Array<GamePlayer> ret(NumPlayers());
--- a/src/games/game.h
+++ b/src/games/game.h
@@ -395,6 +395,7 @@ class GameRep : public BaseGameRep {
   template <class T> friend class TableMixedStrategyProfileRep;
 
 protected:
+  std::vector<GamePlayerRep *> m_players;
   std::string m_title, m_comment;
   unsigned int m_version{0};
 
@@ -535,7 +536,7 @@ public:
   /// Returns the number of players in the game
   virtual size_t NumPlayers() const = 0;
   /// Returns the pl'th player in the game
-  virtual GamePlayer GetPlayer(int pl) const = 0;
+  GamePlayer GetPlayer(int pl) const;
   /// Returns the set of players in the game
   virtual Array<GamePlayer> GetPlayers() const;
   /// Returns the chance (nature) player
--- a/src/games/gameagg.cc
+++ b/src/games/gameagg.cc
@@ -201,20 +201,20 @@ Game GameAGGRep::Copy() const
 Array<int> GameAGGRep::NumStrategies() const
 {
   Array<int> ns;
-  for (int pl = 1; pl <= aggPtr->getNumPlayers(); pl++) {
-    ns.push_back(m_players[pl]->NumStrategies());
+  for (const auto &player : m_players) {
+    ns.push_back(player->NumStrategies());
   }
   return ns;
 }
 
 GameStrategy GameAGGRep::GetStrategy(int p_index) const
 {
-  for (int pl = 1; pl <= aggPtr->getNumPlayers(); pl++) {
-    if (static_cast<int>(m_players[pl]->NumStrategies()) >= p_index) {
-      return m_players[pl]->GetStrategy(p_index);
+  for (const auto &player : m_players) {
+    if (static_cast<int>(player->NumStrategies()) >= p_index) {
+      return player->GetStrategy(p_index);
     }
     else {
-      p_index -= m_players[pl]->NumStrategies();
+      p_index -= player->NumStrategies();
     }
   }
   throw IndexException();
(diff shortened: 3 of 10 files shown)


src/games/game.cc
```diff
@@ -200,6 +200,14 @@ size_t GamePlayerRep::NumSequences() const
 //                            class GameRep
 //========================================================================
 
+GamePlayer GameRep::GetPlayer(int pl) const
+{
+  if (pl <= 0 || pl > m_players.size()) {
```
tturocy, line 205: If you use `.at()` on  a `std::vector` it already does the range checking for you and raises `std::out_of_range` if the index is not valid.

src/games/game.cc
```diff
@@ -200,6 +200,14 @@ size_t GamePlayerRep::NumSequences() const
...
+    throw std::out_of_range("Player index out of range");
+  }
+  return m_players[pl - 1];
```
tturocy, line 208: Because we only need `.at()` this is a one-liner and can be inlined in the class definition.

src/games/game.h
```diff
@@ -438,6 +439,9 @@ class GameRep : public BaseGameRep {
   /// substantive change to the game (i.e. not merely involving labels)
   unsigned int GetVersion() const { return m_version; }
 
+  /// Returns the pl'th player in the game
```
tturocy, line 442: It would be good to keep this function in its original location to avoid unnecessary diffs.

src/games/gameagg.cc
```diff
@@ -202,19 +202,19 @@ Array<int> GameAGGRep::NumStrategies() const
 {
   Array<int> ns;
   for (int pl = 1; pl <= aggPtr->getNumPlayers(); pl++) {
-    ns.push_back(m_players[pl]->NumStrategies());
```
tturocy, line 205: Here is a good opportunity to modernise code by avoiding the use of indices altogether:

```
for (const auto &player : m_players) {
  ns.push_back(player->NumStrategies());
}
```

src/games/gameagg.cc
```diff
@@ -202,19 +202,19 @@ Array<int> GameAGGRep::NumStrategies() const
...
+    ns.push_back(m_players[pl - 1]->NumStrategies());
   }
   return ns;
 }
 
 GameStrategy GameAGGRep::GetStrategy(int p_index) const
 {
   for (int pl = 1; pl <= aggPtr->getNumPlayers(); pl++) {
```
tturocy, line 212: Can also be written as a range-based for-loop (as above example)

src/games/gamebagg.cc
```diff
@@ -213,7 +213,7 @@ GameBAGGRep::GameBAGGRep(std::shared_ptr<agg::BAGG> _baggPtr)
   for (int pl = 1; pl <= baggPtr->getNumPlayers(); pl++) {
     for (int j = 0; j < baggPtr->getNumTypes(pl - 1); j++, k++) {
       m_players.push_back(new GamePlayerRep(this, k, baggPtr->getNumActions(pl - 1, j)));
-      m_players[k]->m_label = lexical_cast<std::string>(k);
```
tturocy, line 216: We can avoid using the index here by using `m_players.back()` instead.

src/games/gamebagg.cc
```diff
@@ -213,7 +213,7 @@ GameBAGGRep::GameBAGGRep(std::shared_ptr<agg::BAGG> _baggPtr)
...
+      m_players[k - 1]->m_label = lexical_cast<std::string>(k);
```
tturocy, line 216: Our `lexical_cast` functions are old compatibility code.  C++ now has `std::to_string` which we prefer when possible.

src/games/gamebagg.cc
```diff
@@ -237,7 +237,7 @@ Array<int> GameBAGGRep::NumStrategies() const
 {
   Array<int> ns;
   for (size_t pl = 1; pl <= NumPlayers(); pl++) {
```
tturocy, line 239: Can just use range-based for loop

src/games/gamebagg.cc
```diff
@@ -246,7 +246,7 @@ int GameBAGGRep::MixedProfileLength() const
 {
   int res = 0;
   for (size_t pl = 1; pl <= NumPlayers(); pl++) {
```
tturocy, line 248: Range-based for loop - or arguably even better `std::aggregate` to use an STL algorithm.

src/games/gameexpl.cc
```diff
@@ -97,7 +97,7 @@ Array<int> GameExplicitRep::NumStrategies() const
   const_cast<GameExplicitRep *>(this)->BuildComputedValues();
   Array<int> dim(m_players.size());
   for (size_t pl = 1; pl <= m_players.size(); pl++) {
-    dim[pl] = m_players[pl]->m_strategies.size();
```
tturocy, line 100: Can be re-written as a range-based for-loop and use `.push_back()` on the returning array.

tturocy (changes requested):
There are some opportunities here for taking care of some C++ modernisation - including ways to get rid of a lot of the index-based expressions altogether.

src/games/gamebagg.cc
d-kad: I went for the range-based option as the alternative code (using `std::accumulate`) looked too bulky and complex

tturocy:
Merged at 58f66f5

PR #524, opened 2025-05-22
Issue 522

### LINKED ISSUE #522: Compute the powers of actions based on the images of nodes immediately descended from actions

d-kad:
Implement, for a given action, its power

This provides computation of the set of plays (terminal nodes) consistent with
passing through a specified action and the plays that do not pass through the information set where this action is available:
* `GameTreeRep::GetPower` in C++
* Property `Action.power` in Python.

Closes #522.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -10,8 +10,11 @@
   been removed as planned. (#357)
 
 ### Added
-- Implement `GetPlays()` (C++) and `get_plays` (Python) to compute the set of terminal nodes consistent
-  with a node, information set, or action (#517)
+- Implement `GameRep::GetPlays` (C++) and `Node.plays`, `Infoset.plays`, and `Action.plays` (Python)
+  to compute the set of terminal nodes consistent with a node, information set, or action (#517)
+- Implement `GameRep::GetPower` (C++) and `Action.power` (Python) to compute the power of an action,
+  that is, the set of plays (terminal nodes) consistent with passing throughn the action
+  and the plays that do not pass through the information set where this action is available.
 
 
 ## [16.3.1] - unreleased
--- a/doc/pygambit.api.rst
+++ b/doc/pygambit.api.rst
@@ -169,6 +169,7 @@ Information about the game
    Action.precedes
    Action.prob
    Action.plays
+   Action.power
 
 .. autosummary::
 
--- a/src/games/gametree.cc
+++ b/src/games/gametree.cc
@@ -23,6 +23,7 @@
 #include <iostream>
 #include <algorithm>
 #include <numeric>
+#include <set>
 
 #include "gambit.h"
 #include "gametree.h"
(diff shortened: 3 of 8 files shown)


ChangeLog
```diff
@@ -10,8 +10,11 @@
   been removed as planned. (#357)
 
 ### Added
-- Implement `GetPlays()` (C++) and `get_plays` (Python) to compute the set of terminal nodes consistent
-  with a node, information set, or action (#517)
+- Implement `GameRep::GetPlays` (C++) and `Node.plays`, `Infoset.plays`, and `Action.plays` (Python)
+  to compute the set of terminal nodes consistent with a node, information set, or action (#517)
+- Implement `GameRep::GetPower` (C++) and `Action.power` (Python) to compute the power of an action,
+  that is, the set of plays (terminal nodes) consistent with passing throughn the action
+  and the plays that do not pass through the information set where this action is available.
```
tturocy: I would suggest just saying 'the power of an action'.  Don't forget the issue number as well.

src/games/gametree.cc
```diff
@@ -1058,6 +1058,24 @@ std::vector<GameNode> GameTreeRep::GetPlays(GameAction action) const
...
 
+std::vector<GameNode> GameTreeRep::GetPower(GameAction action) const
+{
+  std::vector<GameNode> power;
+
+  const std::vector<GameNode> plays = GetPlays(GetRoot());
+  const std::vector<GameNode> aplays = GetPlays(action);
+  const std::vector<GameNode> iplays = GetPlays(action->GetInfoset());
+
+  for (const auto &node : plays) {
```
tturocy: We can just have `GetPlays(GetRoot())` as the target of the iteration directly

src/games/gametree.cc
```diff
@@ -1058,6 +1058,24 @@ std::vector<GameNode> GameTreeRep::GetPlays(GameAction action) const
...
+    if ((std::find(iplays.cbegin(), iplays.cend(), node) == iplays.cend()) ||
```
tturocy: We have a convenience function `contains` defined (in `src/core/util.h`, which is already included) which allows a more idiomatic way of writing these.  (Eventually there will be `contains` in ranges when we someday are at C++23 but in the meanwhile we have the generic function available.)

doc/pygambit.api.rst
```diff
@@ -169,6 +169,7 @@ Information about the game
    Action.precedes
    Action.prob
    Action.plays
+   Action.power
```
tturocy, line 172: This should be `Action.veto`

src/pygambit/action.pxi
```diff
@@ -109,9 +109,18 @@ class Action:
 
     @property
     def plays(self) -> typing.List[Node]:
-        """Returns a list of all terminal `Node` objects consistent with it.
+        """Returns a list of all terminal `Node` objects consistent with the action.
```
tturocy, line 112: Simpler is just to say 'terminal nodes' rather than 'terminal Node objects'

src/pygambit/action.pxi
```diff
@@ -109,9 +109,18 @@ class Action:
...
         """
         return [
             Node.wrap(n) for n in
             self.action.deref().GetInfoset().deref().GetGame().deref().GetPlays(self.action)
         ]
+
```
tturocy, line 118: Docstring has not been updated.

It would be useful to write a brief explanation of what the 'veto' of an action is here - a few lines so the documentation is self-contained

tests/test_actions.py
```diff
@@ -143,3 +143,18 @@ def test_action_plays():
     }  # paths=[0, 1, 0], [0, 1]
 
     assert set(test_action.plays) == expected_set_of_plays
+
+
+def test_action_veto():
```
tturocy, line 148: This explanation would be better put in the docstring of veto rather than in the tests.

tests/test_actions.py
```diff
@@ -143,3 +143,18 @@ def test_action_plays():
...
+    """Verify `action.veto` returns the action's veto
```
tturocy, line 149: It would be useful to have a few test fixtures rather than a single test (see other examples throughout the test suite for setting up fixtures).  In particular think about potential edge cases - e.g. actions at the root node, actions at singleton information sets versus nontrivial information sets, so on.

tturocy (changes requested):

tturocy:
Through separate discussions events have overtaken this idea.  The problem is that whether we look at powers or vetoes, explicitly creating lists or sets of nodes is going to be prohibitively expensive for even the simplest application - so these functions would be of little use to include as core functionality.

PR #528, opened 2025-06-10
Get the action prescribed by a strategy at an information set

d-kad:
This provides an interface `GameStrategyRep::GetAction` to the mapping from infoset to action as part of a strategy for an extensive game.

A function `.action()` has been added in Python as well.

### DIFF
--- a/src/games/game.cc
+++ b/src/games/game.cc
@@ -47,6 +47,26 @@ GameOutcomeRep::GameOutcomeRep(GameRep *p_game, int p_number) : m_game(p_game), 
 }
 
 //========================================================================
+//                      class GameStrategyRep
+//========================================================================
+
+GameAction GameStrategyRep::GetAction(const GameInfoset &p_infoset) const
+{
+  if (p_infoset->GetPlayer() != m_player) {
+    throw MismatchException();
+  }
+  int action = m_behav[p_infoset->GetNumber()];
+  return (action) ? *std::next(p_infoset->GetActions().cbegin(), action - 1) : nullptr;
+}
+
+void GameStrategyRep::DeleteStrategy()
+{
+  if (m_player->GetGame()->IsTree()) {
+    throw UndefinedException();
+  }
+}
+
+//========================================================================
 //                       class GamePlayerRep
 //========================================================================
 
--- a/src/pygambit/gambit.pxd
+++ b/src/pygambit/gambit.pxd
@@ -97,6 +97,8 @@ cdef extern from "games/game.h":
         c_GamePlayer GetPlayer() except +
         string GetLabel() except +
         void SetLabel(string) except +
+        c_GameAction GetAction(c_GameInfoset) except +
+        void DeleteStrategy() except +
 
     cdef cppclass c_GameActionRep "GameActionRep":
         int GetNumber() except +
--- a/src/pygambit/strategy.pxi
+++ b/src/pygambit/strategy.pxi
@@ -73,3 +73,6 @@ class Strategy:
     def number(self) -> int:
         """The number of the strategy."""
         return self.strategy.deref().GetNumber() - 1
+
+    def action(self, infoset: Infoset) -> Action:
+        return Action.wrap(self.strategy.deref().GetAction(infoset.infoset))
--- a/src/games/game.h
+++ b/src/games/game.h
@@ -358,6 +358,12 @@ public:
   GamePlayer GetPlayer() const;
   /// Returns the index of the strategy for its player
   int GetNumber() const { return m_number; }
+
+  /// Returns the action specified by the strategy at the information set
+  GameAction GetAction(const GameInfoset &) const;
+
+  /// Remove this strategy from the game
+  void DeleteStrategy();
   //@}
 };
 


src/pygambit/strategy.pxi
```diff
@@ -73,3 +73,6 @@ class Strategy:
     def number(self) -> int:
         """The number of the strategy."""
         return self.strategy.deref().GetNumber() - 1
+
+    def action(self, infoset: Infoset) -> Action:
```
tturocy: This should use the `_resolve_infoset` method on a game because we allow information sets to be referenced by labels.  This also checks that the information set is from the same game as the action.

tturocy: We need a docstring for this.

src/games/game.cc
```diff
@@ -46,6 +46,26 @@ GameOutcomeRep::GameOutcomeRep(GameRep *p_game, int p_number) : m_game(p_game),
...
+GameAction GameStrategyRep::GetAction(const GameInfoset &p_infoset) const
+{
+  if (p_infoset->GetPlayer() != m_player) {
+    throw MismatchException();
+  }
+  int action = m_behav[p_infoset->GetNumber()];
+  return (action) ? *std::next(p_infoset->GetActions().cbegin(), action - 1) : nullptr;
+}
+
+void GameStrategyRep::DeleteStrategy()
```
tturocy: Why are we adding `DeleteStrategy` here?

tturocy (changes requested):
In addition to the inline notes, we also need an appropriate ChangeLog entry and the function needs adding to the documentation index in the appropriate place

src/pygambit/strategy.pxi
```diff
@@ -73,3 +73,14 @@ class Strategy:
...
         """The number of the strategy."""
         return self.strategy.deref().GetNumber() - 1
+
+    def action(self, infoset: typing.Union[Infoset, str]) -> Action:
+        """Get the action prescribed by a strategy for a given information set.
+
+        Raises KeyError if the strategy is not defined for this information set.
+        """
+        resolved_infoset: Infoset = self.game._resolve_infoset(infoset, "Strategy.action")
+        action: c_GameAction = self.strategy.deref().GetAction(resolved_infoset.infoset)
```
tturocy, line 115: Generally our convention for this situation would be to return `None` rather than throwing an exception.

tests/test_strategies.py
```diff
@@ -0,0 +1,29 @@
+import pytest
```
tturocy: It'd be useful to put this in an existing file rather than adding a new one for just two tests

tests/test_strategies.py
```diff
@@ -0,0 +1,29 @@
...
+
+from . import games
+
+
+def test_strategy_action_defined():
```
tturocy: Let's work on using fixtures to set these up, because also we really should do more than just to test one action for one strategy.  That will also help when we expand this to the bigger tests we want to do for reduced strategy sets

src/pygambit/gambit.pxd
```diff
@@ -97,6 +98,8 @@ cdef extern from "games/game.h":
         c_GamePlayer GetPlayer() except +
         string GetLabel() except +
         void SetLabel(string) except +
+        c_GameAction GetAction(c_GameInfoset) except +
+        void DeleteStrategy() except +
```
tturocy: Wrapping of DeleteStrategy is still here

tests/test_actions.py
```diff
@@ -139,7 +139,57 @@ def test_action_plays():
...
+
+@pytest.fixture(scope="module")
+def e02_game():
+    """Fixture to load the 'e02.efg' game.
+    """
+    return games.read_from_file("e02.efg")
+
+
+@pytest.mark.parametrize(
+    "player_ind, str_ind, infoset_ind, expected_action_ind",
```
tturocy: This is still only testing for one game though; it would be useful to have more than one game for a test.

tests/test_actions.py
```diff
@@ -139,7 +139,57 @@ def test_action_plays():
...
+def test_strategy_action_undefined_returns_none(e02_game, player_ind, str_ind, infoset_ind):
+    """Verify `Strategy.action` returns None when called on an unreached player's infoset
+    """
+    player = e02_game.players[player_ind]
+    strategy = player.strategies[str_ind]
+    infoset = e02_game.infosets[infoset_ind]
+
+    prescribed_action = strategy.action(infoset)
+
+    assert prescribed_action is None
```
tturocy, line 203: There are other error cases to test:

* What happens if the information set belongs to a different player than the strategy?
* What happens if the strategy is in a game with a strategic representation rather than an underlying tree?

src/pygambit/strategy.pxi
```diff
@@ -73,3 +73,15 @@ class Strategy:
     def number(self) -> int:
         """The number of the strategy."""
         return self.strategy.deref().GetNumber() - 1
+
+    def action(self, infoset: typing.Union[Infoset, str]) -> Action:
+        """Get the action prescribed by a strategy for a given information set.
+
+        Returns he prescribed action, or ``None`` if the strategy is not defined
+        for this information set.
+        """
```
tturocy, line 100: Should use numpydoc style:

* Use a versionadded directive to indicate this is new (https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-versionadded)
* Use a Returns section to format the return value
* Use a Raises section to indicate possible exceptions (see notes below about testing error cases)

tturocy (changes requested):

tturocy (approved):

PR #531, opened 2025-06-24
Implement iteration over nodes in DFT order; expose it to Cython

### LINKED ISSUE #530: Implement iteration over nodes in depth-first traversal order
There are some algorithms (e.g. the information set graph mentioned in #484) where we want to iterate over the nodes of the game using depth-first traversal.

Implementation a `Nodes` class as a member of a `GameRep`.  This class should implement an `iterator` that starts from the root node and successively returns nodes in depth-first traversal order - i.e. the root node first, then the first child of the root node (if any), then the first child of that node (if any), and so on.

Then, in Python, there is currently a class which implements this same idea, including the depth-first search done as a recursive generator.  The python implementation can be replaced by a wrapper on the C++ implementation.

Note that the implementation of the iterator in C++ won't be recursive - the most straightforward implementation would be to have a list of iterators over child nodes which functions as a stack.

Tests in Python can be updated accordingly as required.

d-kad:
Introduced a new nested class, ``GameRep::Nodes``, which acts as a standard C++ forward iterator.

- This iterator performs a depth-first, pre-order traversal of the game tree, starting from the root.
- The ``GameRep`` class is now directly iterable, through the addition of: ``begin()`` and ``end()`` methods 
- const overloads for both methods to ensure const-correctness in functions like ``IsPerfectRecall()``.

Closes #530

### DIFF
--- a/src/games/game.h
+++ b/src/games/game.h
@@ -62,7 +62,6 @@ template <class P, class T> class ElementCollection {
 
 public:
   class iterator {
-    friend class GameRep;
 
     P m_owner{nullptr};
     const std::vector<T *> *m_container{nullptr};
@@ -106,6 +105,7 @@ public:
       return *this;
     }
     value_type operator*() const { return m_container->at(m_index); }
+    P GetOwner() const { return m_owner; }
   };
 
   ElementCollection() = default;
@@ -121,6 +121,8 @@ public:
   {
     return m_owner == p_other.m_owner && m_container == p_other.m_container;
   }
+
+  bool empty() const { return size() == 0; }
   size_t size() const { return m_container->size(); }
   GameObjectPtr<T> front() const { return m_container->front(); }
   GameObjectPtr<T> back() const { return m_container->back(); }
@@ -525,7 +527,7 @@ public:
 
       Game m_owner{nullptr};
       GameNode m_current_node{nullptr};
-      std::stack<ChildIterator> m_stack{};
+      std::stack<std::pair<ChildIterator, ChildIterator>> m_stack{};
 
       iterator(Game game) : m_owner(game) {}
 
@@ -561,15 +563,16 @@ public:
         }
 
         if (!m_current_node->IsTerminal()) {
-          m_stack.push(m_current_node->GetChildren().begin());
+          auto children = m_current_node->GetChildren();
+          m_stack.emplace(children.begin(), children.end());
         }
 
         while (!m_stack.empty()) {
-          auto &top_it = m_stack.top();
-          auto end_it = top_it.m_owner->GetChildren().end();
-          if (top_it != end_it) {
-            m_current_node = *top_it;
-            ++top_it;
+          auto &[current_it, end_it] = m_stack.top();
+
+          if (current_it != end_it) {
+            m_current_node = *current_it;
+            ++current_it;
             return *this;
           }
           m_stack.pop();
--- a/tests/test_node.py
+++ b/tests/test_node.py
@@ -803,15 +803,11 @@ def test_node_plays():
     ],
 )
 def test_nodes_iteration_order(game_obj: gbt.Game):
-    """
-    Verify that the C++ `game.nodes` iterator produces the DFS traversal.
-
+    """Verify that the C++ `game.nodes` iterator produces the DFS traversal.
     """
     def dfs(node: gbt.Node) -> typing.Iterator[gbt.Node]:
         yield node
         for child in node.children:
             yield from dfs(child)
 
-    zipped_nodes = itertools.zip_longest(game_obj.nodes, dfs(game_obj.root))
-
-    assert all(a == b for a, b in zipped_nodes)
+    assert all(a == b for a, b in itertools.zip_longest(game_obj.nodes, dfs(game_obj.root)))


src/games/game.h
```diff
@@ -512,6 +513,72 @@ class GameRep : public BaseGameRep {
...
+    }
+
+    Nodes &operator++()
+    {
+      if (m_stack.empty()) {
+        throw std::out_of_range("Cannot increment an end iterator");
+      }
+      const GameNode n = m_stack.top();
+      m_stack.pop();
+      auto children = n->GetChildren();
```
tturocy: This is really inefficient.  We should keep a stack of iterators, not nodes.  Otherwise, we are doing a linear-time operation at each node.

src/pygambit/game.pxi
```diff
@@ -184,13 +186,26 @@ class GameNodes:
         return self.game.deref().NumNodes()
 
     def __iter__(self) -> typing.Iterator[Node]:
-        def dfs(node):
-            yield node
-            for child in node.children:
-                yield from dfs(child)
+        """
```
tturocy: Docstrings are for user consumption.  They shouldn't provide implementation details (unless for some reason that's important for the user); here it is not.

src/pygambit/game.pxi
```diff
@@ -184,13 +186,26 @@ class GameNodes:
...
+        A generator that efficiently iterates over the game nodes using
+        the underlying C++ iterator, without using cdef for local variables.
+        """
         if not self.game.deref().IsTree():
             return
-        yield from dfs(Node.wrap(self.game.deref().GetRoot()))
+
+        # Assign the C++ iterators to regular Python variables.
```
tturocy: There's not a need to write comments that just says what the code does.

tests/test_node.py
```diff
@@ -787,3 +787,33 @@ def test_node_plays():
...
 
     assert set(test_node.plays) == expected_set_of_plays
+
+
+def test_nodes_iteration_matches_python_dfs():
+    """
+    Verify that game.nodes using the C++ iterator produces the same DFS traversal
+    as a pure Python recursive generator.
+    """
+    game = games.read_from_file("e01.efg")
```
tturocy: These should be set up as parameterised tests.

src/games/game.h
```diff
@@ -736,6 +807,10 @@ class GameRep : public BaseGameRep {
   virtual Game SetChanceProbs(const GameInfoset &, const Array<Number> &) = 0;
   //@}
 
+  /// Node iterators
+  Nodes begin() const { return {this, GetRoot()}; }
```
tturocy: These should be defined on the Nodes class, not on Game.

src/games/game.h
```diff
@@ -512,6 +513,76 @@ class GameRep : public BaseGameRep {
   using Players = ElementCollection<Game, GamePlayerRep>;
   using Outcomes = ElementCollection<Game, GameOutcomeRep>;
 
+  class Nodes {
```
tturocy, line 520: This isn't set up properly.

The `Nodes` class is a concept that implicitly represents the set of nodes in a game.

Inside the `Nodes` class should be an `iterator` that does the iteration.

This implementation conflates the two.

Have a look at the implementation of `ElementCollection` to see how this structure works.  The only reason `ElementCollection` doesn't work out of the box is that `ElementCollection` applies when the underlying container is a `std::vector` and we just iterate that.  Here we have a more complex underlying container because the "list" of nodes is actually implemented implicitly as a tree, so we need bespoke iteration.

src/games/game.h
```diff
@@ -512,6 +513,76 @@ class GameRep : public BaseGameRep {
...
+    }
+
+    bool operator==(const Nodes &other) const
+    {
+      return m_owner == other.m_owner && m_current_node == other.m_current_node;
+    }
+
+    bool operator!=(const Nodes &other) const { return !(*this == other); }
+
+    friend class GameRep;
```
tturocy: Our preferred style is to list `friend` classes at the top of the class definition where needed.

src/games/game.h
```diff
@@ -512,6 +513,76 @@ class GameRep : public BaseGameRep {
   using Players = ElementCollection<Game, GamePlayerRep>;
   using Outcomes = ElementCollection<Game, GameOutcomeRep>;
 
+  class Nodes {
+  private:
+    using ChildIterator = ElementCollection<GameNode, GameNodeRep>::iterator;
+
+    Game m_owner{nullptr};
+    GameNode m_current_node{nullptr};
+    std::stack<std::pair<ChildIterator, ChildIterator>> m_stack{};
```
tturocy: Why is it not enough just to have a `std::stack<std::vector<GameNodeRep *>::iterator>` ?

src/pygambit/game.pxi
```diff
@@ -184,13 +186,20 @@ class GameNodes:
         return self.game.deref().NumNodes()
 
     def __iter__(self) -> typing.Iterator[Node]:
-        def dfs(node):
-            yield node
-            for child in node.children:
-                yield from dfs(child)
+        """
+        A generator that efficiently iterates over the game nodes using
```
tturocy: This is still a docstring that focuses on implementation details.  Meanwhile it misses the essential thing about this iterator, which is it is a depth-first traversal.

src/pygambit/game.pxi
```diff
@@ -184,13 +186,20 @@ class GameNodes:
...
+        the underlying C++ iterator, without using cdef for local variables.
+        """
         if not self.game.deref().IsTree():
             return
-        yield from dfs(Node.wrap(self.game.deref().GetRoot()))
+
+        it = self.game.deref().begin()
```
tturocy: Cython will correctly make use of C++ iterators using just a `for` loop syntax - see all the instances where we use `ElementCollection` in Cython (e.g. say in `NodeChildren` at `node.pxi:44`.

d-kad:
@tturocy I implemented node traversal in operator++ with a single children iterator on stack using exception catching mechanism (mirroring your Python code snippet for the PR-checker)

src/games/game.h
```diff
@@ -724,6 +816,8 @@ class GameRep : public BaseGameRep {
   //@{
   /// Returns the root node of the game
   virtual GameNode GetRoot() const = 0;
+  /// Returns the nodes of the game in the DFT order
```
tturocy: The `Nodes` object per se doesn't impose any ordering.  We have just one iterator at the moment which happens to implement DFT.  We could imagine having other iterators which do other traversals.

src/games/game.h
```diff
@@ -512,6 +513,97 @@ class GameRep : public BaseGameRep {
...
+      }
+
+      iterator &operator++()
+      {
+        if (!m_current_node) {
+          throw std::out_of_range("Cannot increment an end iterator");
+        }
+
+        auto children = m_current_node->GetChildren();
+        if (children.size() > 0) {
```
tturocy: `children.empty()` is considered more idiomatic

src/games/game.h
```diff
@@ -512,6 +513,97 @@ class GameRep : public BaseGameRep {
...
+          m_stack.emplace(children.begin());
+        }
+
+        while (!m_stack.empty()) {
+          try {
+            *m_stack.top();
+            break;
+          }
+          catch (std::out_of_range) {
```
tturocy: Exceptions should be used only for, well, exceptions.  Exception handling is actually rather expensive, so should only be used in situations where something unexpected has happened (and so performance is less critical).  The more idiomatic solution is to compare against the appropriate end iterator for the container.

As an aside, it's considered better to catch by reference rather than value; catching by value results in a copy (and loses information in the event of derived classes).  That's moot here but will surely come up elsewhere in future.

src/games/game.h
```diff
@@ -512,6 +513,97 @@ class GameRep : public BaseGameRep {
...
+    };
+
+    /// Default constructor to support declaration by other modules (e.g. Cython)
+    Nodes() = default;
+
+    /// Constructor for the Nodes range.
+    explicit Nodes(Game p_owner) : m_owner(p_owner) {}
+
+    /// Returns an iterator to the first node (the root).
+    iterator begin() const { return m_owner ? iterator(m_owner, m_owner->GetRoot()) : iterator(); }
```
tturocy: You should be able to replace these with braced initialiser lists so you don't have to repeat the class name.

We prefer always to put parentheses around the boolean condition in the ternary operator.

src/pygambit/game.pxi
```diff
@@ -28,6 +28,8 @@ import scipy.stats
 
 import pygambit.gameiter
 
+from cython.operator cimport dereference, preincrement
```
tturocy: We don't use these imports anymore

tests/test_node.py
```diff
@@ -787,3 +787,29 @@ def test_node_plays():
...
+        pytest.param(games.read_from_file("poker.efg")),
+        pytest.param(gbt.Game.new_tree()),
+    ],
+)
+def test_nodes_iteration_order(game_obj: gbt.Game):
+    """
+    Verify that the C++ `game.nodes` iterator produces the DFS traversal.
+
+    """
+    def dfs(node: gbt.Node) -> typing.Iterator[gbt.Node]:
```
tturocy, line 808: This technically returns a `Generator`

tests/test_node.py
```diff
@@ -787,3 +787,29 @@ def test_node_plays():
...
+        yield node
+        for child in node.children:
+            yield from dfs(child)
+    nodes_iter = list(game_obj.nodes)
```
tturocy: Coverting generators to lists throws away the value of having generators!

The best way to do this would be something like

```
assert all(a == b for a, b in zip(game_obj.nodes, dfs(game_obj.root), strict=True)
```

This never builds the lists in memory (so this would work for quite huge games) while also failing immediately at the first disagreement between the generators.

tturocy (changes requested):
As this looks to be converging well towards a solution please do squash on the next iteration and give a good candidate commit message.

tests/test_node.py
```diff
@@ -787,3 +788,30 @@ def test_node_plays():
...
+    """
+    Verify that the C++ `game.nodes` iterator produces the DFS traversal.
+
+    """
+    def dfs(node: gbt.Node) -> typing.Iterator[gbt.Node]:
+        yield node
+        for child in node.children:
+            yield from dfs(child)
+
+    zipped_nodes = itertools.zip_longest(game_obj.nodes, dfs(game_obj.root))
```
tturocy: I'd suggest it's slightly nicer to just do this in one line rather than assigning to a variable that's used just once.

tests/test_node.py
```diff
@@ -787,3 +788,30 @@ def test_node_plays():
...
+        pytest.param(games.read_from_file("cent3.efg")),
+        pytest.param(games.read_from_file("e01.efg")),
+        pytest.param(games.read_from_file("e02.efg")),
+        pytest.param(games.read_from_file("poker.efg")),
+        pytest.param(gbt.Game.new_tree()),
+    ],
+)
+def test_nodes_iteration_order(game_obj: gbt.Game):
+    """
+    Verify that the C++ `game.nodes` iterator produces the DFS traversal.
```
tturocy: Slightly cleaner docstring style is to do this here on one line but still use the triple-quotes as in `__iter__` above.  Certainly avoid the trailing blank line.

tturocy (changes requested):

src/games/game.h
```diff
@@ -566,7 +568,7 @@ class GameRep : public BaseGameRep {
 
         while (!m_stack.empty()) {
           auto &top_it = m_stack.top();
-          auto end_it = top_it.m_owner->GetChildren().end();
+          auto end_it = top_it.GetOwner()->GetChildren().end();
```
tturocy: An even more straightforward way to do this would be for the iterator stack to be a `std::stack<std::pair<ChildIterator, ChildIterator>>`, where we just capture the end iterator straightaway as the second element of the pair, and don't have to figure it out every step.

tturocy (changes requested):

PR #532, opened 2025-06-24
Refactor ``GameTreeRep::IsPerfectRecall`` with efficient single-pass algorithm; add new tests

### LINKED ISSUE #484: ENH: Improve implementation of `IsPerfectRecall`
Our implementation of `GameTreeRep::IsPerfectRecall` is quite inelegant and unnecessarily inefficient.

* Rewrite the C++ implementation.  Most likely use of `GameInfosetRep::Precedes(GameNode)` and iteration over STL containers (rather than index-based iteration) will help a lot.
* Add appropriate test cases to the `pygambit` test suite to exercise this, for both general imperfect recall examples and absent-mindedness.

d-kad:
This pull request completely refactors the old ``GameTreeRep::IsPerfectRecall`` method

- Replaced the old, nested-loop implementation with a linear-time, single-pass DFT algorithm. 
- The new implementation correctly handles games with Chance moves by initializing the history for the Chance player. This resolves a ``map::at key-not-found error`` that occurred when traversing games like Kuhn Poker.
- Added a new, parameterized test suite covering both perfect and imperfect recall cases (namely, forgetting past action, forgetting past information, absent-mindedness) 

Closes #484


src/games/gametree.cc
```diff
@@ -732,51 +733,52 @@ bool GameTreeRep::IsConstSum() const
...
+  // Histories track the last actions of players taken to reach a given node
+  using History = std::map<GamePlayer, std::optional<Assignment>>;
 
-          if (i == j && precedes) {
-            s1 = iset1;
-            s2 = iset2;
-            return false;
-          }
+  std::map<GameInfoset, std::optional<Assignment>> last_actions_for_infoset;
+  std::map<GameNode, History> last_actions_for_node;
```
tturocy: This is inefficient especially for a deep tree, which is the case we care about most.  We should just have a list of actions for each player prior to the current node in the iteration; that list can be appended to or the last action removed as we iterate.

src/games/gametree.cc
```diff
@@ -732,51 +733,52 @@ bool GameTreeRep::IsConstSum() const
...
+
+    if (!inserted) {
+      if (it->second != player_last_action) {
+        if (!player->IsChance()) {
+          return false;
         }
       }
     }
+
+    for (const GameAction action : infoset->GetActions()) {
```
tturocy: Iterating over the actions/children here defeats the purpose of iterating over the nodes.  Because this now means it's not a single pass; we are iterating each node twice.

tturocy (changes requested):

src/games/game.h
```diff
@@ -484,6 +504,76 @@ class GameNodeRep : public GameObject {
   bool IsSubgameRoot() const;
 };
 
+class ActionsIterator {
```
tturocy: This should be a class inside `GameNodeRep::Actions` and not in the (Gambit) global namespace.  You can define it using the fully-qualified name e.g. `class GameNodeRep::Actions::iterator` at top level.

src/games/game.cc
```diff
@@ -188,6 +188,20 @@ size_t GamePlayerRep::NumSequences() const
                                [](const GameInfosetRep *s) { return s->m_actions.size(); });
 }
 
+//========================================================================
+//                            class GameNodeRep
+//========================================================================
+
+GameNodeRep::Actions::iterator GameNodeRep::Actions::begin() const
```
tturocy: Define these `inline` in the header.  Making them actual function calls will take an unnecessary performance hit.

src/games/gametree.cc
```diff
@@ -905,6 +866,43 @@ std::vector<GameNodeRep *> GameTreeRep::BuildConsistentPlaysRecursiveImpl(GameNo
...
+    ActionsIterator &current_it = position.top();
+    const GameNode parent = current_it.GetOwner();
+
+    if (current_it != parent->GetActions().end()) {
+      auto [action, child] = *current_it;
+
+      prior_actions[parent->GetPlayer()].top() = action;
+
+      if (!child->IsTerminal()) {
+        m_infosetParents[child->GetInfoset()].push_back(prior_actions[child->GetPlayer()].top());
```
tturocy: We really want `m_infosetParents` not to contain duplicate actions.  Otherwise we are basically just making a second copy of the tree in memory, and in any event then have to reduce the set of parent actions to do anything with it.

src/games/gametree.cc
```diff
@@ -736,61 +736,21 @@ bool GameTreeRep::IsConstSum() const
...
-    }
-
-    else {
-      prior_actions[parent->GetPlayer()].pop();
-      position.pop();
-    }
-  }
-
-  for (const auto &[infoset, parent_action_options] : infoset_parents) {
+  for (const auto &[infoset, parent_action_options] : m_infosetParents) {
```
tturocy: With `m_infosetParents` written to reduce the set of actions already, then this can be written more succinctly using `std::any_of`.

src/games/game.h
```diff
@@ -105,6 +105,7 @@ template <class P, class T> class ElementCollection {
       return *this;
     }
     value_type operator*() const { return m_container->at(m_index); }
+    P GetOwner() const { return m_owner; }
```
tturocy: This could return `const P &` instead.

src/games/game.h
```diff
@@ -458,6 +459,23 @@ class GameNodeRep : public GameObject {
...
+  public:
+    class iterator;
+
+    explicit Actions(GameNode p_owner);
+
+    iterator begin() const;
+    iterator end() const;
+  };
+
+  /// @brief Returns a range for iterating over this node's (action, child) pairs.
```
tturocy: It is better to call this a "collection" and not a "range".  Especially as later C++ now has a concept of "range" which is not exactly what we have here.

src/games/game.h
```diff
@@ -488,6 +506,100 @@ class GameNodeRep : public GameObject {
...
+public:
+  /// @name Lifecycle
+  //@{
+  /// Default constructor. Creates an iterator in a past-the-end state.
+  iterator() = default;
+
+  /// Creates a new iterator that zips an action iterator and a child iterator.
+  iterator(GameInfosetRep::Actions::iterator p_action_it,
+           GameNodeRep::Children::iterator p_child_it);
+  //   : m_action_it(p_action_it), m_child_it(p_child_it)
```
tturocy: If we are going to have this inlined later then we need to remove the commented code.

src/games/gametree.cc
```diff
@@ -732,54 +734,20 @@ bool GameTreeRep::IsConstSum() const
...
-          }
-        }
-      }
-    }
+  if (GetRoot()->IsTerminal()) {
+    return true;
   }
 
-  return true;
+  const bool violation = std::any_of(m_infosetParents.cbegin(), m_infosetParents.cend(),
```
tturocy: We can just return this directly without assigning to a variable (and could use `std::all_of` instead to avoid having to do the negation)

src/games/gametree.h
```diff
@@ -39,6 +39,7 @@ class GameTreeRep : public GameExplicitRep {
   std::size_t m_numNodes = 1;
   std::size_t m_numNonterminalNodes = 0;
   std::map<GameNodeRep *, std::vector<GameNodeRep *>> m_nodePlays;
+  std::map<GameInfoset, std::set<GameAction>> m_infosetParents;
```
tturocy: Looking ahead to other changes being made in parallel with respect to ownership and memory management semantics on game objects, would be better to use `GameInfosetRep *` and `GameActionRep *` here.  Should not have a substantial effect on other code.

src/games/gametree.cc
```diff
@@ -893,6 +862,43 @@ std::vector<GameNodeRep *> GameTreeRep::BuildConsistentPlaysRecursiveImpl(GameNo
   return consistent_plays;
 }
 
+void GameTreeRep::BuildInfosetParents()
+{
+  std::map<GamePlayer, std::stack<GameAction>> prior_actions;
+  std::stack<GameNodeRep::Actions::iterator> position;
+
+  for (auto player : GetPlayers()) {
```
tturocy: In this function, we've got access directly to the underlying data structure.  So we can be writing e.g. `for (const auto &player : m_players` here, and likewise accessing data members directly rather than via the public calls.  The public calls trigger reference counting/memory management accounting which isn't useful here (and just adds a bit of overhead)

src/games/gametree.cc
```diff
@@ -893,6 +860,43 @@ std::vector<GameNodeRep *> GameTreeRep::BuildConsistentPlaysRecursiveImpl(GameNo
...
 
+void GameTreeRep::BuildInfosetParents()
+{
+  std::map<GamePlayer, std::stack<GameAction>> prior_actions;
+  std::stack<GameNodeRep::Actions::iterator> position;
+
+  for (auto player : m_players) {
+    prior_actions[player].emplace(nullptr);
+  }
+  prior_actions[GetChance()].emplace(nullptr);
```
tturocy: We should use direct member access rather than the smart-pointer wrappers, as the latter creates overhead.  So for example this can just be `m_chance` instead of `GetChance()`.  Likewise elsewhere.

src/games/gametree.cc
```diff
@@ -893,6 +860,43 @@ std::vector<GameNodeRep *> GameTreeRep::BuildConsistentPlaysRecursiveImpl(GameNo
...
+  while (!position.empty()) {
+    GameNodeRep::Actions::iterator &current_it = position.top();
+    const GameNode parent = current_it.GetOwner();
+
+    if (current_it != parent->GetActions().end()) {
+      auto [action, child] = *current_it;
+
+      prior_actions[parent->GetPlayer()].top() = action;
+
+      if (!child->IsTerminal()) {
```
tturocy: Here we also need to check whether this information set has been visited on the path and if so ensure we take the same action as previously taken, per our discussion.

src/games/gametree.cc
```diff
@@ -893,6 +891,82 @@ std::vector<GameNodeRep *> GameTreeRep::BuildConsistentPlaysRecursiveImpl(GameNo
...
+    return;
+  }
+
+  for (auto player : m_players) {
+    prior_actions[player].emplace(nullptr);
+  }
+  prior_actions[m_chance].emplace(nullptr);
+
+  position.emplace(m_root->GetActions().begin());
+  prior_actions[m_root->GetPlayer()].emplace(nullptr);
```
tturocy: Should be `m_root->m_player`

src/games/gametree.cc
```diff
@@ -893,6 +891,82 @@ std::vector<GameNodeRep *> GameTreeRep::BuildConsistentPlaysRecursiveImpl(GameNo
...
+  m_infosetParents[m_root->GetInfoset()].insert(nullptr);
```
tturocy: Should be `m_root->m_infoset`

src/games/gametree.cc
```diff
@@ -893,6 +891,82 @@ std::vector<GameNodeRep *> GameTreeRep::BuildConsistentPlaysRecursiveImpl(GameNo
...
+
+  while (!position.empty()) {
+    auto &current_it = position.top();
+    auto parent = current_it.GetOwner();
+
+    if (current_it != parent->GetActions().end()) {
+      auto [action, child] = *current_it;
+
+      prior_actions[parent->GetPlayer()].top() = action;
```
tturocy: Should be `parent->m_player`

src/games/gametree.cc
```diff
@@ -893,6 +891,82 @@ std::vector<GameNodeRep *> GameTreeRep::BuildConsistentPlaysRecursiveImpl(GameNo
...
+      initial_choice[parent->GetInfoset()] = {parent, action};
```
tturocy: Should be `parent->m_infoset`

src/games/gametree.cc
```diff
@@ -893,6 +891,82 @@ std::vector<GameNodeRep *> GameTreeRep::BuildConsistentPlaysRecursiveImpl(GameNo
...
+      auto initial_choice_it = initial_choice.find(child->GetInfoset());
+      while (initial_choice_it != initial_choice.end()) {
+        auto initial_action = initial_choice_it->second.second;
+        auto prior_action_ff = prior_actions[child->GetPlayer()].top();
+        m_infosetParents[child->GetInfoset()].insert(prior_action_ff);
+
+        auto newchild = child->GetChild(initial_action);
+
+        prior_actions[child->GetPlayer()].emplace(initial_action);
+        fast_forward_history.emplace_back(child->GetPlayer());
```
tturocy: Should be `child->m_player`.  (And similar with all other calls to public member functions, these create unnecessary reference counting.)

src/games/gametree.cc
```diff
@@ -893,6 +891,82 @@ std::vector<GameNodeRep *> GameTreeRep::BuildConsistentPlaysRecursiveImpl(GameNo
...
+
+      // records every emplace made onto the prior_actions stack during a fast-forward.
+      std::vector<GamePlayer> fast_forward_history;
+
+      // fast forward absent-minded child nodes
+      auto initial_choice_it = initial_choice.find(child->GetInfoset());
+      while (initial_choice_it != initial_choice.end()) {
+        auto initial_action = initial_choice_it->second.second;
+        auto prior_action_ff = prior_actions[child->GetPlayer()].top();
+        m_infosetParents[child->GetInfoset()].insert(prior_action_ff);
```
tturocy: Should be `child->m_infoset`

src/games/gametree.cc
```diff
@@ -893,6 +891,82 @@ std::vector<GameNodeRep *> GameTreeRep::BuildConsistentPlaysRecursiveImpl(GameNo
...
+      initial_choice[parent->GetInfoset()] = {parent, action};
+
+      // records every emplace made onto the prior_actions stack during a fast-forward.
+      std::vector<GamePlayer> fast_forward_history;
+
+      // fast forward absent-minded child nodes
+      auto initial_choice_it = initial_choice.find(child->GetInfoset());
+      while (initial_choice_it != initial_choice.end()) {
+        auto initial_action = initial_choice_it->second.second;
+        auto prior_action_ff = prior_actions[child->GetPlayer()].top();
```
tturocy: Should be `child->m_player`

src/games/gametree.cc
```diff
@@ -893,6 +891,82 @@ std::vector<GameNodeRep *> GameTreeRep::BuildConsistentPlaysRecursiveImpl(GameNo
...
+    auto parent = current_it.GetOwner();
+
+    if (current_it != parent->GetActions().end()) {
+      auto [action, child] = *current_it;
+
+      prior_actions[parent->GetPlayer()].top() = action;
+      initial_choice[parent->GetInfoset()] = {parent, action};
+
+      // records every emplace made onto the prior_actions stack during a fast-forward.
+      std::vector<GamePlayer> fast_forward_history;
```
tturocy: I suggest that the way to represent these is by defining the `position` to be something like

```
using AbsentMindedEdge = std::pair<GameActionRep *, GameNodeRep *>;
using ActiveEdge = std::variant<GameNodeRep::Actions::iterator, AbsentMindedEdge>;

std::stack<ActiveEdge> position;
```

If not `std::variant` then some other form of union.  The `using` aliases are just suggestive, don't know if they are the best names for them, just wanting to avoid having thrice-nested template class names.

We'll still need code to determine which type of step is taken at the top of the `position` stack, but this keeps the information all in one place and is a lot easier to understand.g

tturocy (changes requested):

PR #534, opened 2025-06-27
Issue 533

d-kad:
Two steps:

1. Added `GameTreeRep::BuildPlayerConsequences`; stored `m_playerConsequences` as a private member variable of the `GameTreeRep` class and make it a part of `BuildComputedValues`;
2. Added debug logs to `GameTreeRep::BuildPlayerConsequences` to illustrate the novel structure, those will be eventually removed;
3. TODO: implement `GameTreeRep::MakeReducedStratPR`.


tturocy (changes requested):
Some organisational things:
* This should be proposed to be merged back into the issue branch, not master.
* PRs should to the extent possible be separate from each other.  This one does not depend on the revised implementation of the perfect recall checker, so that should not be being included here.

This actually does not do anything with computing the reduced strategies, which is the whole point of this exercise.  And it's the data that that algorithm needs which determines whether the player consequences data structure that's being built here is the right one (or not).

PR #535, opened 2025-07-14
typo: node -> infoset in _resolve_infoset

rahulsavani:


tturocy:
This has been a bug since 16.2, so we should do the PR against maint16_2.

tturocy (changes requested):
This is fine, but we need to have this against the branch maint16_2 instead of master.

PR #540, opened 2025-08-05
Readthedocs build instructions

edwardchalstrey1:
TODO:
- [x] Add instructions for pushing the changes to readthedocs, since this may currently be a manual step
    - @tturocy @rahulsavani how is this currently done? This would be good to automate - see https://github.com/alan-turing-institute/gambit_admin/issues/2

### DIFF
--- a/doc/build.rst
+++ b/doc/build.rst
@@ -144,3 +144,20 @@ using `nose2`.
 
 Once installed, simply ``import pygambit`` in your Python shell or
 script to get started.
+
+Editing the documentation
+--------------------------
+
+Devcontainer version:
+
+1. Open the Repo in VS Code which should automatically give you the option to open in a devcontainer.
+2. Open a terminal in VS Code.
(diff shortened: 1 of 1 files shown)


tturocy:
This is configured on the readthedocs side, so those builds are automated once we've pushed to GitHub.  But that only works for branches we've activated.  So testing the build online for proposed changes isn't as straightforward - one can of course build locally but that requires installing sphinx and in any event sometimes the local builds work and then the online one doesn't because of environment differences.

edwardchalstrey1:
> This is configured on the readthedocs side, so those builds are automated once we've pushed to GitHub. But that only works for branches we've activated. So testing the build online for proposed changes isn't as straightforward - one can of course build locally but that requires installing sphinx and in any event sometimes the local builds work and then the online one doesn't because of environment differences.

Great, thanks for explaining, I've opened #543 to look into automating this, in the meantime I think we can merge these local build instructions that worked for me

PR #566, opened 2025-09-18
Add GUI install instructions to docs

edwardchalstrey1:
- [x] mac
- [ ] Windows
- [ ] Linux

### DIFF
--- a/doc/gui.rst
+++ b/doc/gui.rst
@@ -25,58 +25,70 @@ To build larger games or to explore parameter spaces of a game
 systematically, it is recommended to use
 :ref:`the Python package <pygambit>`.
 
-Install on Mac
-==============
+Installation Instructions
+=========================
 
-To build and install the Gambit GUI on macOS, follow these steps:
+.. tab-set::
 
-1. **Install build dependencies:**
+   .. tab-item:: macOS
 
-   .. code-block:: bash
+      To build and install the Gambit GUI on macOS, follow these steps:
 
-      brew install automake autoconf libtool
+      1. **Install build dependencies:**
 
-   .. note::
-      If you encounter interpreter errors with autom4te, you may need to ensure
-      your Perl installation is correct or reinstall the autotools:
+         .. code-block:: bash
 
-      .. code-block:: bash
+            brew install automake autoconf libtool
 
-         brew reinstall automake autoconf libtool
+         .. note::
+            If you encounter interpreter errors with autom4te, you may need to ensure
+            your Perl installation is correct or reinstall the autotools:
 
-2. **Download and build wxWidgets:**
+            .. code-block:: bash
 
-   .. code-block:: bash
+               brew reinstall automake autoconf libtool
 
-      curl -L -O https://github.com/wxWidgets/wxWidgets/releases/download/v3.2.8/wxWidgets-3.2.8.tar.bz2
-      tar xjf wxWidgets-3.2.8.tar.bz2
-      cd wxWidgets-3.2.8
-      mkdir build-release
-      cd build-release
-      ../configure --disable-shared --disable-sys-libs
-      make -j4
-      sudo make install
+      2. **Download and build wxWidgets:**
 
-3. **Build and install Gambit:**
+         .. code-block:: bash
 
-   Navigate back to the Gambit source directory and run:
+            curl -L -O https://github.com/wxWidgets/wxWidgets/releases/download/v3.2.8/wxWidgets-3.2.8.tar.bz2
+            tar xjf wxWidgets-3.2.8.tar.bz2
+            cd wxWidgets-3.2.8
+            mkdir build-release
+            cd build-release
+            ../configure --disable-shared --disable-sys-libs
+            make -j4
+            sudo make install
 
-   .. code-block:: bash
+      3. **Build and install Gambit:**
 
-      aclocal
-      automake --add-missing
-      autoconf
-      ./configure
-      make
-      sudo make install
+         Navigate back to the Gambit source directory and run:
 
-4. **Create macOS application bundle:**
+         .. code-block:: bash
 
-   To create a distributable DMG file:
+            aclocal
+            automake --add-missing
+            autoconf
+            ./configure
+            make
+            sudo make install
 
-   .. code-block:: bash
+      4. **Create macOS application bundle:**
 
-      make osx-dmg
+         To create a distributable DMG file:
+
+         .. code-block:: bash
+
+            make osx-dmg
+
+   .. tab-item:: Windows
+
+      TODO: Add Windows installation instructions
+
+   .. tab-item:: Linux
+
+      TODO: Add Linux installation instructions
 
 .. toctree::
    :maxdepth: 2
--- a/doc/gui.rst
+++ b/doc/gui.rst
@@ -25,6 +25,58 @@ To build larger games or to explore parameter spaces of a game
 systematically, it is recommended to use
 :ref:`the Python package <pygambit>`.
 
+Install on Mac
+==============
+
+To build and install the Gambit GUI on macOS, follow these steps:
+
+1. **Install build dependencies:**
+
+   .. code-block:: bash
+
+      brew install automake autoconf libtool
+
+   .. note::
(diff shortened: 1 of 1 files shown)


edwardchalstrey1:
@rahulsavani @tturocy Is the GUI installable on Windows/Linux or only Mac?

tturocy:
There are .msi installers for Windows built with every release.

edwardchalstrey1:
> There are .msi installers for Windows built with every release.

Thanks Ted, I have changed it to link to the releases page where the installers can be found. I left the mac manual build instructions in since unfortunately my Turing Mac couldn't trust the dmg downloaded from the release page wasn't malware - so I guess others might find this too

tturocy:
Apple makes it deliberately misleading as to how to install/run unsigned apps.  Meanwhile they charge an extortionate amount to sign them (at least from the perspective of an open-source developer).  Does Turing have some dev account that can be used to sign these?

Building on Windows is at best going to be complicated, because at present it is going to require having e.g. MinGW installed so as to use the autotools setup, followed by having to have wxWidgets installed correctly as well.  While it would be good to document this, equally in truth the documentation for "building Gambit on Windows" would consist of "(1) Throw out your computer and get one that runs MacOS or Linux; (2) See those build directions"  It would be easier! :)

edwardchalstrey1:
@tturocy @rahulsavani I've had a good go at updating the installation pages today. Let me know your thoughts, here are the key pages with changes:
- [Docs homepage](https://gambitproject--566.org.readthedocs.build/en/566/) now has a clear link to an install page and a quickstart instruction to install pygambit
- New [Install page](https://gambitproject--566.org.readthedocs.build/en/566/install.html) expains PyGambit install and a few different ways to install the CLI/GUI (this page is linked to at the top of the PyGambit, GUI and CLI pages as well)
- [Building Gambit from source](https://gambitproject--566.org.readthedocs.build/en/566/developer.build.html) in the developer docs has also been revised with sequential steps and platform specific parts

The main thing that's lacking are the Windows sections as we discussed, but I think if anything the way I've presented it especially in the dev docs gets across that we haven't tested these recently. We can update these parts in a subsequent pull request.

doc/developer.build.rst
```diff
@@ -1,143 +1,162 @@
...
 
-Un*x users, please note that Gambit at this time only supports the
-GTK port of wxWidgets.
+.. note::
+  If you don't want to build the graphical interface, you can pass the argument `--disable-gui` to the configure step, for example, ::
 
-If wxWidgets it isn't installed in a standard place (e.g., /usr or
-/usr/local), you'll need to tell configure where to find it with the
---with-wx-prefix=PREFIX option, for example::
+  `./configure --disable-gui`
```
tturocy: These backticks shouldn't be there (and in a few other places below) - because they make it look like you need to type them to run the command.

doc/developer.build.rst
```diff
@@ -1,143 +1,162 @@
...
-Building the extension follows the standard approach.
-From the **root directory of the source tree** execute
+in the Gambit source tree. We recommend to install `pygambit`
+as part of a virtual environment rather than in the system's Python.
+Use `pip` to install from the **root directory of the source tree**, optionally including the `-e` flag for an editable install:
+
+.. code-block:: bash
 
-    python -m pip install .
+   pip install -e .
```
tturocy: Editable installs are not very useful because they do not automatically re-build the compiled/Cython portions of the package.  They only affect the pure Python parts, which are only a very small part of it.

doc/developer.contributing.rst
```diff
@@ -22,12 +22,20 @@ When reporting a bug, please be sure to include the following:
...
+
 Contributing code
 ----------------
 
 Gambit is an open-source project, and contributions are welcome from anyone.
 The project is hosted on GitHub, and contributions can be made via pull requests following the standard GitHub workflow.
 
+In the git repository, the branch ``master`` always points to the
+latest development version.  New development should in general always
+be based off this branch.  Branches labeled ``maintVV``, where ``VV``
```
tturocy: Update this perhaps to ``maintX_Y`` where `X` is the major version number and `Y` the minor version number?

doc/developer.contributing.rst
```diff
@@ -22,12 +22,20 @@ When reporting a bug, please be sure to include the following:
...
+is the version number, point to the latest commit on a stable
```
tturocy: The term "stable version" is a bit of a holdover from the way we used to version things (we used to have a bit more in the way of "development" releases which weren't guaranteed to be "stable"). What's a more modern term for what these releases that come off of previous versions would be called?

doc/developer.build.rst
```diff
@@ -1,143 +1,162 @@
...
-version.
+      .. note::
+         Note that Gambit requires wxWidgets version 3.2.x or higher.
 
+.. dropdown:: Install on Windows
+   :class-container: sd-border-0
 
-For Windows users
------------------
+    .. warning::
```
tturocy: I would suggest we tell people they need to install MSYS/MinGW and then follow the Unix-style instructions.  This is implemented in the `tools` workflow in Github actions and is used to build the GUI installer - so clearly that *does* work!!!!

doc/developer.contributing.rst
edwardchalstrey1: Questions:
- For patch releases e.g. 16.3.1 do you still use the `maint16_3` branch? If so can mention that here.
- Is the plan is there are no intentionally unstable releases going forward?

At the moment, someone wanting to access experimental features can install from master branch, and see docs by clicking "latest" - but perhaps if you want to also release bleeding edge changes, but not have them be there by default for new users to avoid instability we can set up Readthedocs (& PyPI) to point "stable" to an earlier release than the most recent one, if that was desirable?

Screenshot 2025-11-10 at 10 41 50

doc/developer.build.rst
edwardchalstrey1: Thanks @tturocy , I have refactored [that page](https://gambitproject--566.org.readthedocs.build/en/566/developer.build.html) including adding Windows instructions based on the GH action steps - I don't actually have access to a Windows machine to test, so can you let me know if that looks right to you now?

doc/developer.contributing.rst
```diff
@@ -22,12 +22,19 @@ When reporting a bug, please be sure to include the following:
...
+
 Contributing code
 ----------------
 
 Gambit is an open-source project, and contributions are welcome from anyone.
 The project is hosted on GitHub, and contributions can be made via pull requests following the standard GitHub workflow.
 
+In the git repository, the branch ``master`` always points to the
+latest development version.  New development should in general always
+be based off this branch.  Branches labeled ``maintX_Y``, where ``X`` is the major version number and ``Y`` is the minor version number, point to the latest commit on a stable
```
edwardchalstrey1, line 35: @tturocy are you happy with this addition that we can delete: https://github.com/gambitproject/gambit/wiki/Workflow

doc/developer.contributing.rst
tturocy: Yes, `maint16_3` contains all of the micro versions of 16.3.

I wouldn't foresee releasing unstable bleeding-edge versions.  Anyone wanting to try out such a feature would need to use either `master` or a topic branch for a feature that's under development.  We do build binary artefacts so someone could well do that without having to set up a development environment.  But at least at the moment I don't think we would have any plans to advertise these publicly - this would be the sort of thing we'd tell someone about 1-to-1 if they had a need of an in-development feature.

doc/developer.build.rst
tturocy: I made a few edits.  The `make` target works fine.  We don't use it in the Github action because we need to set the environment variable differently for it to find the WiX toolset there.

doc/developer.contributing.rst
tturocy: Yes.  I have disabled the wiki now.

tturocy:
Is there anything else pending, or should we merge these as it's already a chunky set of changes and iterate further as we go.

edwardchalstrey1:
> Is there anything else pending, or should we merge these as it's already a chunky set of changes and iterate further as we go.

No I think this is good to go!

PR #573, opened 2025-10-07
Add Gambit/OpenSpiel interoperability tutorial

edwardchalstrey1:
Closes #555 
Includes changes from #621 

See the tutorial [here](https://gambitproject--573.org.readthedocs.build/en/573/tutorials/interoperability_tutorials/openspiel.html)

and where it fits on the PyGambit docs pages [here](https://gambitproject--573.org.readthedocs.build/en/573/pygambit.html#interoperability-tutorials)

### Changes
- Adds OpenSpiel tutorial to a new "Interoperability tutorials" section
- Updates local running tutorial instructions to include OpenSpiel
- Also adds a few items to `.gitignore`

Issue #552, opened 2025-08-28
Add easy way to visualise game trees in PyGambit


edwardchalstrey1:
### This feature request is for...

PyGambit

### Requested enahancement or feature

When working on #542 I noticed there doesn't seem to be a way to display the Game trees for extensive form games with Python.

For normal form games, after you have set up a game with `Game.new_table` and then added payoffs, there is a nice feature in a Jupyter notebook where running a cell with the `pygambit.gambit.Game` object displays the payout matrix pandas-style:

<img width="327" height="277" alt="Image" src="https://github.com/user-attachments/assets/3ea6a57f-179d-4369-9aac-4192b3a85874" />

It would be very useful for the tutorials and in general to be able to quickly view the shape of your extensive form `Game` object set up with `Game.new_tree` with a display function, which could similarly get triggered when just running a cell with the game object. You could see the tree grow as you add nodes. 

### Implementation

It might make sense to use an existing library, perhaps we could convert a Gambit game to [NetworkX](https://networkx.org/documentation/stable/tutorial.html#drawing-graphs) to make use of that plotting functionality.

### Drawbacks

I guess this may end up being quite a big task as it could entail re-implementing a lot of the visualisation features already present in the GUI. That being said, I can imagine this would be very handy especially for viewing large extensive form games with modern Python plotting libraries.

@tturocy @rahulsavani @d-kad Is this something you have already discussed, or did I miss a way to do this that already exists? For me, when learning the PyGambit API in a notebook context exploring setting up a game, it feels difficult not being able to see the tree I'm building as I go

edwardchalstrey1:
### September 17th 2025 update

I set up one card poker in GTE and downloaded the `.ef` to run with Bernhard's code:

GTE version            |  drawTree
:-------------------------:|:-------------------------:

<img width="861" height="801" alt="Image" src="https://github.com/user-attachments/assets/76590543-0f84-4503-95a6-c440a137aa48" />  |  <img width="654" height="443" alt="Image" src="https://github.com/user-attachments/assets/3f8e48bd-38b9-4eb5-946b-d7f74055f632" />

PR #576, opened 2025-10-27
[WIP] Integrate drawtree

edwardchalstrey1:
See todo list in #552 

Pre-requisite PR:
- #573 

See: https://gambitproject--576.org.readthedocs.build/en/576/tutorials/03_poker.html

### DIFF
--- a/.github/ISSUE_TEMPLATE/tutorial_request.yml
+++ b/.github/ISSUE_TEMPLATE/tutorial_request.yml
@@ -3,19 +3,25 @@ description: Submit a request for a new tutorial to be added to the Gambit docum
 title: "[Tutorial]: "
 labels: ["tutorial", "documentation"]
 body:
+  - type: markdown
+    attributes:
+      value: |
+        Thanks for taking the time to fill out this tutorial request!
   - type: textarea
     id: describe
     attributes:
       label: Describe the tutorial you want to see.
       description: Go into as much detail as you can regarding the tutorial you would like to see, how long it should be etc.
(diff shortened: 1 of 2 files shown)


edwardchalstrey1:
### draw tree todos

Best approach would be to pass a game object from pygambit to draw_tree and whilst pygambit is not a requirement of draw_tree (since we don't need it to draw .ef files that already have layout encoded), using the gambit layout functionality does require it and throws an error if pygambit is not installed.

- [ ] Move `DefaultLayout` to a separate .py file from core
- [ ] Rename `efg_to_ef` to reflect that this uses the "default layout" not Gambit layout
    - Later on, might be better to remove this completely and if an EFG is passed in, use the gambit layout, via first reading the game with gambit 
- [ ] Add `gambit_to_ef` function that accepts a `pygambit.gambit.Game` object and creates the layout tree dict with `gbt.layout_tree(g)` and uses these to create an ef representation (we need label, payoff etc info from the game as well as the layout). The `import pygambit as gbt` will need to happen within this function as we cannot expect all users of draw_tree to have it installed.
- [ ] Update `generate_tikz`, `draw_tree` and `generate_X` functions to accept a gambit game object as an alternative to ef file. Within `generate_tikz` it should check the object type and if a gambit game object is passed in, use the new `gambit_to_ef` function. You can then with a game `g` have the code in the tutorial simply be `draw_tree(g)`.
- [ ] Do the above on a new branch in draw_tree and increment the version number

PR #579, opened 2025-10-31
Absent-Mindedness Checker

### LINKED ISSUE #578: ENH: Add function returning whether an information set is absent-minded
Absent-mindedness occurs when a player, moving along a single path of play, can re-enter an information set visited previously. Detecting this property is valuable for users to:

- Better understand the structure of their game.
- Be warned if using certain concepts and solution algorithms that are incompatible with absent-mindedness.

The existing `GameTreeRep::BuildInfosetParents()` method already contains the necessary traversal logic to detect this condition.

1. Add a new data member to `GameTreeRep`: `std::set<GameInfoset> m_absentMindedInfosets`;
2. Augment the existing `BuildInfosetParents()` traversal to populate `m_absentMindedInfosets` that tracks all absent-minded information sets.
3. A new public method, `IsAbsentMinded() const`, will mirror the logic of `IsPerfectRecall()`: it will trigger the traversal if needed and then check the state of the new container.

d-kad:
Closes #578 

This PR implements detection of absent-mindedness in extensive-form games, where a player can re-enter an information set on a single path of play.

Changes:

- Augments the `BuildInfosetParents()` traversal to detect and record absent-minded information sets.
- Adds `IsAbsentMinded()` in C++ and a corresponding `is_absent_minded` property in Python, mirroring the logic of `IsPerfectRecall()`.
- Includes a new test suite to verify the implementation.
- Updates the ChangeLog and api.rst.

### DIFF
--- a/tests/test_extensive.py
+++ b/tests/test_extensive.py
@@ -50,6 +50,55 @@ def test_game_add_players_nolabel():
 
 
 @pytest.mark.parametrize("game_input,expected_result", [
+    # Games without absent-mindedness (includes all games with perfect recall)
+    ("e01.efg", False),
+    ("e02.efg", False),
+    ("cent3.efg", False),
+    ("poker.efg", False),
+    ("basic_extensive_game.efg", False),
+    ("wichardt.efg", False),  # forgetting past action; Wichardt (GEB, 2008)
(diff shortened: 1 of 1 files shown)


tturocy:
Per off-line discussion, we're going to make absent-mindedness a property of information sets rather than games; changes for this are in progress.

PR #586, opened 2025-11-06
tests for EFG Nash solvers -- enumpoly, lp, lcp -- in behavior strategies

rahulsavani:


tests/games.py
```diff
@@ -18,6 +18,32 @@ def read_from_file(fn: str) -> gbt.Game:
...
+    g = gbt.Game.new_tree(players=["1", "2"], title=title)
+    actions1 = [str(i) for i in range(m)]
+    actions2 = [str(i) for i in range(n)]
+    g.append_move(g.root, "1", actions1)
+    for node in g.root.children:
+        g.append_move(node, "2", actions2)
+    iset = g.root.children[0].infoset
+    for c in g.root.children:
+        g.set_infoset(c, iset)
+    from itertools import product
```
tturocy: PEP 8 seems to imply that imports should be global and at the top of the file, and not inside of functions.  I'd also weakly prefer just `import itertools` and using `itertools.product` just for the avoidance of possible name clashes.

tests/games.py
```diff
@@ -18,6 +18,32 @@ def read_from_file(fn: str) -> gbt.Game:
         raise ValueError(f"Unknown file extension in {fn}")
 
 
+def create_efg_corresponding_to_bimatrix_game(
+    A: np.ndarray, B: np.ndarray, title: str
+) -> gbt.Game:
+    """
+    There is not direct pygambit method to create an EFG from a stategic-form game.
```
tturocy: It could be interesting functionality possibly to have some transformation that does this.  But we would have to think a bit more about how it might be implemented, so probably just fine to leave it as an auxiliary function here.

tests/games.py
```diff
@@ -18,6 +18,32 @@ def read_from_file(fn: str) -> gbt.Game:
...
+    Here we create an EFG corresponding to a bimatrix game, which is given by two numpy arrays.
+    Player 1 moves first.
+    """
+    assert A.shape == B.shape
+    m, n = A.shape
+    g = gbt.Game.new_tree(players=["1", "2"], title=title)
+    actions1 = [str(i) for i in range(m)]
+    actions2 = [str(i) for i in range(n)]
+    g.append_move(g.root, "1", actions1)
```
tturocy, line 34: `append_move` takes a list of nodes, all of which are then placed in the same information set, specifically to avoid having to do this kind of awkward manipulation!

https://gambitproject.readthedocs.io/en/latest/api/pygambit.gambit.Game.append_move.html#pygambit.gambit.Game.append_move

tests/games.py
```diff
@@ -104,6 +142,128 @@ def create_myerson_2_card_poker_efg() -> gbt.Game:
...
+        g.append_move(g.root.children[i].children[0], "Bob", ["Check", "Bet"])
+    # If Bob bets after Alice checked (Bob's second action), then Alice can Fold or Call.
+    # Append Alice's response nodes for the check-then-bet branch
+    for i in range(6):
+        g.append_move(g.root.children[i].children[0].children[1], "Alice", ["Fold", "Call"])
+    # If Alice bets initially, Bob can Fold or Call.
+    for i in range(6):
+        g.append_move(g.root.children[i].children[1], "Bob", ["Fold", "Call"])
+    # Set up information sets to represent imperfect information.
+    # Alice dealt J
```
tturocy: See comment above.  Shouldn't have to use `set_infoset` when building a game from scratch if you're taking full advantage of `append_move` and tracking histories.

tests/games.py
```diff
@@ -104,6 +142,128 @@ def create_myerson_2_card_poker_efg() -> gbt.Game:
...
+    # Alice dealt K
+    g.set_infoset(g.root.children[4].children[0].children[1],
+                  g.root.children[5].children[0].children[1].infoset)
+    # Add outcomes at terminal nodes (net payoffs); first define four outcomes:
+    alice_wins1 = g.add_outcome([1, -1], label="Alice wins 1")
+    alice_wins2 = g.add_outcome([2, -2], label="Alice wins 2")
+    bob_wins1 = g.add_outcome([-1, 1], label="Bob wins 1")
+    bob_wins2 = g.add_outcome([-2, 2], label="Bob wins 2")
+    # Check-Check -> Bob wins 1
+    g.set_outcome(g.root.children[0].children[0].children[0], bob_wins1)
```
tturocy: I would suggest writing a utility function that takes a sequence of action labels and then finds the appropriate node.  This ought to be core functionality (ideas for such path notation have been floated before).  But it would make the code much cleaner to write tuples like `("A-K, "B-Q", "C", "C")` for Alice gets king, Bob gets queen, both check, and have a function that traverses the tree to find the right node.

Then you could just write the payoff table as a dict from such tuples to outcomes, anda apply them in a loop.

tests/test_nash.py
```diff
@@ -57,20 +143,22 @@ def test_lcp_strategy_double():
...
+        data=[
+            [gbt.Rational(1, 3), gbt.Rational(2, 3), gbt.Rational(0), gbt.Rational(0)],
+            [gbt.Rational(2, 3), gbt.Rational(1, 3)],
+        ],
     )
     assert result.equilibria[0] == expected
 
 
-def test_lcp_behavior_double():
+def test_lcp_behavior_strategy_double():
```
tturocy: We call these "mixed behaviors" to parallel "mixed strategies"

tturocy (changes requested):
See comments.  The game-building can be expressed *much* more felicitously.  And indeed should we consider building Kuhn poker also to a more prominent example in the documentation somewhere?  It's far more engaging than the one-card "stripped down" poker (although it poses some practical challenges as we have been discussing).  That could be spun off into a separate discussion.

tests/games.py
rahulsavani: Agreed.

tests/games.py
rahulsavani: Indeed, thanks.

tests/games.py
rahulsavani: The Kuhn poker code came from GameInterpreter v2 and was much more verbose before I took a pass last night. I agree that it is a nice example and it makes sense to create it in the best way now in the tests, and consider having it as an example elsewhere. At least for now, LLMs will certainly need extra information on how to create nice pygambit code, but probably not a priority as major changes are likely in the not too distant future anyway on the pygambit side.

tturocy: Yes, Kuhn poker is important enough it's worth doing well.  And also, part of the reason GameInterpreter generates spaghetti code is in part that we don't have a high-quality training dataset.  So just putting slop in the test suite would just compound the problem. ;)

tests/games.py
rahulsavani: OK, I have gone a different route for now: since the actual relevant sequences of actions are already encoded in the tree, and the tree is not too large, I got at these action sequences by getting the set of terminal nodes, and constructing their paths back to the route in terms of action labels. Hopefully you find this reasonable enough for now.

tests/test_nash.py
rahulsavani: I see. I made these changes because I didn't like the fact that "mixed strategies" where being referred to simply as "strategies", so I added the "mixed" in front. Then, rather than adding "mixed" before behavior, I added "strategy" after. I will just revert to the original naming for now (strategy = mixed strategy; behavior = mixed behavior strategy).

tturocy: Fair enough, but in the interests of pedantry (and who isn't) I've tried to discipline myself to say "mixed behaviour" and not "mixed behaviour strategy", just to restrict the word "strategy" to the normal form sense.  I know that many people do say "behaviour strategy", but this does lead to the confusion as above!

rahulsavani: Understood.

Issue #587, opened 2025-11-07
FEAT: New syntax for node children and paths


tturocy:
Constructing paths in an extensive game is currently quite tedious and verbose.  To this end the following features are proposed:

1.   Indexing into the children of a node will be done by *action label* rather than the label of the child node.  This is a breaking change but should be seen as definitely an improvement (see e.g. the discussion in #553 for an example where the current behaviour was unanticipated)

2.  Implement the operator `+` with LHS of a node and RHS of an action label producing a node.

PR #588, opened 2025-11-07
New syntax for node children and paths (#587)

tturocy:
Implements idea in #587.

Have added tests, but not yet added examples to user guide pending input on whether we want to do anything slightly different.

There should be some opportunities also to use this in the test suite - e.g. @rahulsavani, not to be re-writing the Kuhn poker example already, but would this syntax make it a bit more clear than what you've done?


rahulsavani:
> @rahulsavani, not to be re-writing the Kuhn poker example already, but would this syntax make it a bit more clear than what you've done?

Yes, thanks, I've re-written it and and it's much nicer now.

PR #590, opened 2025-11-08
weaker test for enumpoly that ignores output order

### LINKED ISSUE #589: TEST: enumpoly test failing on Windows with Python 3.13
See the runs of the GitHub action.  This is a problem with setting the tolerance - a nice example though of the system-dependence of the enumpoly implementation (this isn't the first time we've gotten variation in output on Windows there).

rahulsavani:
closes #589 

The tests for enumpoly_solve in mixed behaviors included a 3-player game where the order of equilibria found varies from run to run on Windows under the github tests action -- as a result, tests would sometimes pass and sometimes fail, since this test expects a pre-defined order of output equilibria.

That old test has been renamed: test_enumpoly_behavior_rational -> test_enumpoly_ordered_behavior_rational, and the new test_enumpoly_unordered_behavior_rational, which checks all found equilibria are in a pre-defined *set*, has been added.

### DIFF
--- a/tests/test_nash.py
+++ b/tests/test_nash.py
@@ -106,20 +106,20 @@ def test_enummixed_rational():
         # ),
     ],
 )
-def test_enumpoly_ordered_behavior_rational(
+def test_enumpoly_ordered_behavior(
     game: gbt.Game, mixed_behav_prof_data: list, stop_after: typing.Union[None, int]
 ):
-    """Test calls of enumpoly for mixed behavior equilibria, rational precision,
+    """Test calls of enumpoly for mixed behavior equilibria,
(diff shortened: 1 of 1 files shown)


rahulsavani:
> I have re-run the relevant Windows test 10 times in total and it never failed. We can do some repeat runs of that on this PR too.

OK, it's passed 10 re-runs since the PR: https://github.com/gambitproject/gambit/actions/runs/19191745733

One thing to note is that in both the ordered and unordered versions, now maxregret is set at 1e-4, rather than using the default (1e-3). The tests will probably repeatedly pass with the default regret, if we prefer to revert to that.

Issue #593, opened 2025-11-10
Move sequence form implementation to games directory, update enumpoly to use


tturocy:
The `enumpoly` solver has an explicit class which represents the sequence form of an extensive game.  In contrast, although `lp` and `lcp` both also use the sequence form, the representations there are implicit.  Therefore we have duplicated (technically, triplicated) code for computations involving the sequence form.

As a first refactoring step, move the existing `GameSequenceForm` and related support classes from `src/solvers/enumpoly` to `src/game`, and make the existing sequence form object an implementation detail by exposing appropriate member functions on `BehaviorSupport`.  It should then be straightforward to revise the implementation of `enumpoly` to call functions on the support rather than the sequence form object directly.

tturocy:
Merged at commit 9995912.

PR #595, opened 2025-11-10
Sequence form

StephenPasteris:
Issue #593

### DIFF
--- a/src/games/behavspt.h
+++ b/src/games/behavspt.h
@@ -30,7 +30,7 @@
 namespace Gambit {
 
 class GameSequenceForm;
-struct GameSequenceRep;
+class GameSequenceRep;
 class SequencesWrapper;
 class PlayerSequencesWrapper;
 class InfosetsWrapper;
--- a/src/games/gameseq.h
+++ b/src/games/gameseq.h
@@ -28,7 +28,7 @@
 
 namespace Gambit {
 
-struct GameSequenceRep {
+class GameSequenceRep {
 public:
   GamePlayer player;
   GameAction action;


Makefile.am
```diff
@@ -285,6 +285,10 @@ agg_SOURCES = \
 
 game_SOURCES = \
 	src/gambit.h \
+	src/games/gameseq.cc \
+	src/games/gameseq.h \
+	src/games/indexproduct.h \
```
tturocy: We can just leave indexproduct.h in enumpoly.

src/games/behavspt.cc
```diff
@@ -259,4 +260,59 @@ std::list<GameNode> BehaviorSupportProfile::GetMembers(const GameInfoset &p_info
...
+{
+  if (!m_sequenceForm) {
+    m_sequenceForm = std::make_shared<GameSequenceForm>(*this);
+  }
+  return m_sequenceForm;
+}
+
+SequencesWrapper BehaviorSupportProfile::GetSequences() const
+{
+  auto sequences = GetSequenceForm()->GetSequences();
```
tturocy: Can we get by with these being one-liners rather than defining the variable?

src/games/behavspt.h
```diff
@@ -29,6 +29,13 @@
 
 namespace Gambit {
 
+class GameSequenceForm;
+struct GameSequenceRep;
```
tturocy: `GameSequenceRep` should be a class (it'll be used as one in later development)

tturocy: The conceptual model for sequences is very similar to that for strategies (in terms of how we represent them, generate them on demand, and so on).  So we want to evolve sequences in that direction.

tturocy: Let's make the `GameSequenceRep` members private now as a step in this direction.

src/games/gameseq.h
```diff
@@ -293,6 +293,61 @@ class GameSequenceForm {
...
+  auto begin() const { return m_sequences.begin(); }
+  auto end() const { return m_sequences.end(); }
+
+  std::size_t size() const { return m_sequences.size(); }
+
+private:
+  GameSequenceForm::PlayerSequences m_sequences;
+};
+
+class InfosetsWrapper {
```
tturocy, line 327: I believe we can do away with this as a wrapper.  Instead let's just define an Infosets collection directly as part of the support (a nested class within the support).  We should only need to define `.size()` for this right now as that is all that is used.  This should return the number of reachable information sets, derived by counting the `true` values in the reachable infosets map.

PR #600, opened 2025-11-11
Implement `Game.get_own_prior_actions`

### LINKED ISSUE #582: ENH: Add function returning the player's immediately preceding actions at an information set
For the analysis of imperfect recall (and also absent mindedness, see #578),  it is useful to know, for a given information set, which are the most recent action(s) of the same player which precede this information set.  For example, this is always a singleton in the case of perfect recall.

Computing these is already implemented internally.  We will expose these as `GameRep::GetOwnPriorActions(GameInfoset)` in C++ and `Game.get_own_prior_actions(infoset)` in Python.  

And as always add appropriate tests to exercise this.  Because these data relate closely to testing imperfect call, tests of this function may help streamline and focus imperfect recall tests.

d-kad:
Implement `GameTreeRep::GetOwnPriorActions` (C++) and `Game.get_own_prior_actions` (Python) to compute, for a given information set, the set of last actions taken by the player acting in the information set before reaching it 

Closes #582

### DIFF
--- a/src/pygambit/game.pxi
+++ b/src/pygambit/game.pxi
@@ -790,6 +790,9 @@ class Game:
         yet played prior to reaching that node, the own prior action is null, which is represented
         by `None` in the list of actions returned.
 
+        If a member node is not reachable due to the path to the node passing through an
+        absent-minded information set, that node has no own prior action.
+
         .. versionadded:: 16.5
 
         Returns


src/pygambit/game.pxi
```diff
@@ -777,6 +777,23 @@ class Game:
...
+        of the player active in it which precede this information set.
+
+        Raises
+        ------
+        RuntimeError
+            If the information set is not reachable from the root of the game.
+        """
+        infoset = self._resolve_infoset(infoset, "get_own_prior_actions")
+
+        own_prior_actions = []
```
tturocy: This can be more compactly written as a list comprehension.

src/pygambit/game.pxi
```diff
@@ -777,6 +777,23 @@ class Game:
         """Whether the game is constant sum."""
         return self.game.deref().IsConstSum()
 
+    def get_own_prior_actions(self, infoset: typing.Union[Infoset, str]):
+        """ For a given information set, find the most recent action(s)
+        of the player active in it which precede this information set.
+
```
tturocy, line 792: We need a bit more clarity here about what an "own_prior_action" is.  Also we need to note it was added in version 16.5.

tturocy (changes requested):
Looking at the code, I am going to go back a bit on what I said in yesterday's discussion on this.

If I am reading correctly, the current implementation does *NOT* include the null action as a prior action.  By my understand of how we are defining this concept, the null action does need to be part of the set of prior actions.  So for example if an information set is always the first one reached by the player (if it is reached), then its `own_prior_actions` would be `[None]`.

We cannot just omit `None` because consider the case of an information set that can be reached either from the root node or via some other action `a` (in a game without perfect recall).  Then the set of `own_prior_actions` must be `[None, a]` (in some order).  Otherwise it would appear this information set does not violate perfect recall (which it does).

Assuming what I have just written is correct, then this also means we can define this function completely well for information sets unreachable via absent-mindedness - they can have the *empty list* of prior actions, which is different from the list `[None]`.

@rahulsavani Do you agree with my analysis, or again am I misunderstanding something?

rahulsavani:
> @rahulsavani Do you agree with my analysis, or again am I misunderstanding something?

This sounds correct to me, and I like that we can just return the empty list for unreachable info sets, and indeed this will be a characterisation of them, right?

PR #608, opened 2025-11-12
remove use of poker.efg in test suite

### LINKED ISSUE #606: TEST: Duplicated game (simple poker) in test suite
There are two games in the test suite which are actually the same game:
* `poker.efg`, loaded by `games.create_1_card_poker_efg`
* `myerson_2_card_poker.efg`, loaded by `create_myerson_2_card_poker_efg`.

The two games have different naming conventions - it's a one-card game in the sense that one card is dealt, but it's a two-card game in the sense there are two possible cards in the deck that can be dealt.

I don't immediately see a reason these should be separate games for testing purposes; we should just have one instance of it.

The games differ in their labels, and also whether they express the game data as floating-point or rational numbers in the representation.  So updating the tests should be straightforward, but would take a bit of time and a bit of attention.

rahulsavani:
resolves #606

### DIFF
--- a/tests/test_games/stripped_down_poker.efg
+++ b/tests/test_games/stripped_down_poker.efg
@@ -1,0 +1,14 @@
+EFG 2 R "A simple Poker game" { "Fred" "Alice" }
+"Stripped-Down Poker: a simple game of one-card poker from Reiley et al (2008)."
+
+c "" 1 "(0,1)" { "King" 1/2 "Queen" 1/2 } 0
+p "" 1 1 "(1,1)" { "Bet" "Fold" } 0
+p "" 2 1 "(2,1)" { "Call" "Fold" } 0
+t "" 1 "Win Big" { 2, -2 }
+t "" 2 "Win" { 1, -1 }
+t "" 4 "Lose" { -1, 1 }
+p "" 1 2 "(1,2)" { "Bet" "Fold" } 0
+p "" 2 1 "(2,1)" { "Call" "Fold" } 0
+t "" 3 "Lose Big" { -2, 2 }
+t "" 2 "Win" { 1, -1 }
+t "" 4 "Lose" { -1, 1 }


tturocy (changes requested):
We should have just one version of this game.  It is at the moment awkward that we have games in `tests/test_games` and in `contrib/games`.  But we will need to sort that out later, and I expect it will gradually come out in the was as we build out tests for algorithms and solution concepts.

We should decide which version we want to have, because there are several issues here:

* We want a version that uses rational fractions rather than decimal representation.  The decimal representation version would be very old (when "floating point" payoffs were permitted).  We want to encourage the use of rational payoffs where it makes sense to do so.

* I believe Myerson (1991) likely gets priority on this game as written, I do not have an earlier citation and there is none in his textbook.  However two issues to consider:
  - Myerson treats Red (high card) followed by Fold as a *win* for the informed player.  This is consistent with the narrative exposition in the book, but is not consistent with how "poker" works.  Alternatively, the version of this game used in Reiley's classroom experiment follows the "normal" rules of poker.
  - Myerson does not actually give the players human names.  However the names Fred and Alice are consistent with his textbook, as he uses male pronouns for odd-numbered players and female for even-numbered ones.

* We also need a name for the game.  Calling this 2-card poker is not accurate.  In Myerson's description, the deals are "red" and "black" because they're dealt from a standard deck of cards but only the suit colour matters for payoffs.  Further, calling it 2-card poker makes it sound more like e.g. Borel or von Neumann's versions, where each player gets a "card" (a draw from the uniform distribution).

Tagging @edwardchalstrey1 because this interacts with the relevant tutorial.

edwardchalstrey1:
> Tagging @edwardchalstrey1 because this interacts with the relevant tutorial.

In the tutorial the efg file isn't used as the game is constructed via the code so not to worry

tturocy:
Yes @edwardchalstrey1, but it would be good that the game we build there matches the saved file, just for the avoidance of any confusion!!! :)

rahulsavani:
@tturocy: The tests now only use stripped_down_poker.efg, which is faithful to Reiley et al in terms of action labels and payoffs. The reuse of action labels ("Fold" and "Bet" in particular) means that a couple of tests were simply dropped, since they relied on the uniqueness of said labels. I don't think that's a problem. I didn't want to drop some other tests that used infoset labels, so I inserted the infoset labels from the old myerson_2_card_poker.efg into stripped_down_poker.efg, which was adapted from poker.efg which had empty infoset labels.

@edwardchalstrey1 : For after this is settled on any merged into the tests: 

contrib/games still contains poker.efg, and tests/test_userguide.py still contains test_myerson_poker(). If you could update both of these when you update the user guide that would be great (putting stripped_down_poker.efg in contrib/games, and, assuming we keep test_userguid.py, updating it to create the right game, exactly as in the updated user guide, not mentioning Myerson at all, but Reiley et al (2008).

edwardchalstrey1:
> assuming we keep test_userguide.py

@rahulsavani I actually dropped this test and replaced it with a new one that checks all tutorials can be run without error, which doesn't need to be updated when any new tutorials are added or existing ones changed, looks like Reiley et al (2008) is cited correctly as well **Edit:** actually it mentions them both, so should I drop the Myerson citation?

edwardchalstrey1:
Also the resultant EFG from tutorial 3 has slightly different labels for players, actions and payoff nodes, and lacks the full title . I propose I should update the tutorial version to have the proper title and standardisation of "Fold" (vs "Pass"), but perhaps better to keep the terminal node labels referencing the player that wins e.g. "Alice wins big" (indifferent over the actual names of course 😄).

Tutorial 3 EFG:

```
EFG 2 R "One card poker" { "Alice" "Bob" }
""

c "" 1 "" { "King" 1/2 "Queen" 1/2 } 0
p "" 1 1 "" { "Raise" "Fold" } 0
p "" 2 1 "" { "Meet" "Pass" } 0
t "" 1 "Alice wins big" { 2, -2 }
t "" 2 "Alice wins" { 1, -1 }
t "" 4 "Bob wins" { -1, 1 }
p "" 1 2 "" { "Raise" "Fold" } 0
p "" 2 1 "" { "Meet" "Pass" } 0
t "" 3 "Bob wins big" { -2, 2 }
t "" 2 "Alice wins" { 1, -1 }
t "" 4 "Bob wins" { -1, 1 }
```

New stripped down poker EFG from this PR:
```
EFG 2 R "A simple Poker game" { "Fred" "Alice" }
"Stripped-Down Poker: a simple game of one-card poker from Reiley et al (2008)."

c "" 1 "(0,1)" { "King" 1/2 "Queen" 1/2 } 0
p "" 1 1 "(1,1)" { "Bet" "Fold" } 0
p "" 2 1 "(2,1)" { "Call" "Fold" } 0
t "" 1 "Win Big" { 2, -2 }
t "" 2 "Win" { 1, -1 }
t "" 4 "Lose" { -1, 1 }
p "" 1 2 "(1,2)" { "Bet" "Fold" } 0
p "" 2 1 "(2,1)" { "Call" "Fold" } 0
t "" 3 "Lose Big" { -2, 2 }
t "" 2 "Win" { 1, -1 }
t "" 4 "Lose" { -1, 1 }
```

edwardchalstrey1:
Also, what is the difference between: `p "" 1 1 "" { "Raise" "Fold" } 0` and `p "" 1 1 "(1,1)" { "Raise" "Fold" } 0`, what does the "(1,1)" do?

tturocy:
> Also, what is the difference between: `p "" 1 1 "" { "Raise" "Fold" } 0` and `p "" 1 1 "(1,1)" { "Raise" "Fold" } 0`, what does the "(1,1)" do?

That field is an explicit information set label.  It is possibly dubious to be labelling these in that way given the broader discussion about how we order information sets in "canonicalisation" - but for present purposes it's good enough.  When we have our standardised games library this will become moot.

rahulsavani:
Ah was typing at the same time as Ted.

> Let's go with "Alice" and "Bob" universally perhaps just to be consistent both internally and with the "folk" naming?

@tturocy , shall I update the test version to use Alice and Bob then?

tturocy:
Maybe "Alice has King", "Alice has Queen", "Bob's response" or something like that?  Wouldn't want to overload "Bob" to mean both a player and an information set.

PR #609, opened 2025-11-13
new tests for enumpure_solve in pure strategies and behaviors

rahulsavani:


rahulsavani:
> We should add them as part of the changes for #617. Also, I note that none of the test instances for enumpure_solve "agent" currently have agent-form equilibria that are not also Nash equilibria -- such tests should be added.

This is now done in this branch with a game now called `myerson_fig_4_2.efg`, which is tested in `test_nash.py::test_enumpure_agent` to ensure that we find both the true Nash and the agent Nash.

Note that `myerson_fig_4_2.efg` is also used in `test_behavior.py::test_agent_max_regret_versus_non_agent` which confirms that while `agent_max_regret` is 0, `max_regret` >0 for the non-true-Nash agent equilibrium.

PR #611, opened 2025-11-13
Add tests to verify the new `SubgameRootFinder` prototype

d-kad:
The added tests verify the new `SubgameRootFinder` Python prototype against the legacy `is_subgame_root` logic.

The new algorithm correctly identifies all legacy subgame roots, and this test suite confirms that any additional roots it finds are exclusively members of absent-minded information sets, which the legacy method was designed to exclude.

### DIFF
--- a/tests/test_node.py
+++ b/tests/test_node.py
@@ -1,5 +1,7 @@
 import itertools
 import typing
+from heapq import heappop, heappush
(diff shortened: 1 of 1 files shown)


tturocy:
Because a fundamentally different approach is needed, closing this one.

PR #613, opened 2025-11-14
Sequence form

StephenPasteris:
Removed the need for wrapper classes in gameseq.h

### DIFF
--- a/src/games/gameseq.cc
+++ b/src/games/gameseq.cc
@@ -43,7 +43,7 @@ void GameSequenceForm::BuildSequences(const GameNode &n,
     auto tmp_sequence = p_currentSequences.at(n->GetPlayer());
     for (auto action : m_support.GetActions(n->GetInfoset())) {
       if (m_correspondence.find(action) == m_correspondence.end()) {
-        m_sequences[n->GetPlayer()].push_back(std::make_shared<GameSequenceRep>(
+        m_sequences[n->GetPlayer()].emplace_back(std::make_shared<GameSequenceRep>(
             n->GetPlayer(), action, m_sequences[n->GetPlayer()].size() + 1,
             tmp_sequence.get_shared()));
         m_correspondence[action] = m_sequences[n->GetPlayer()].back();


src/games/behavspt.cc
```diff
@@ -22,6 +22,7 @@
 
 #include "gambit.h"
 #include "gameseq.h"
+#include "behavsptseqform.h"
```
tturocy: The objective of this exercise is that the sequence form should *not* be a separate object at the end of the day; it is conceptually an integral part of the concept of a Game.  The fact there was this other class hanging around was a product of some dubious design/programming decisions a long time ago.  Our goal is to fix that.

So we definitely do not want to be adding more files.  We are wanting fewer, not more.  Further, doing things this way also defeats the purpose of the exercise, because `enumpoly` (or other methods) are supposed not to worry about implementation details, yet this `#include` ties them more closely to implementation details.

src/games/behavsptseqform.h
```diff
@@ -0,0 +1,174 @@
+//
```
tturocy: As noted elsewhere we're not going to add a new file with this change.  But if and when you do add files, the header comments should follow the conventions as in all other files.

src/games/behavsptseqform.h
```diff
@@ -0,0 +1,174 @@
...
+#define GAMBIT_BEHAVSPT_SEQ_FORM_H
+
+#include "gambit.h"
+#include "gameseq.h"
+
+namespace Gambit {
+
+class BehaviorSupportProfile::Sequences {
+private:
+  const std::shared_ptr<GameSequenceForm> m_sfg;
```
tturocy: The reason you're finding yourself having to create this file for inclusion in `enumpoly` is here.  The private member here needs now to be a `BehaviorSupportProfile`, not the sequence form object.  Change this in all of these classes (and the contractors, &c) and the need to expose this as a separate header file goes away.

src/games/behavsptseqform.h
```diff
@@ -0,0 +1,174 @@
...
+  {
+    return m_sfg->m_sequences.at(m_player).begin();
+  }
+  std::vector<GameSequence>::const_iterator end() const
+  {
+    return m_sfg->m_sequences.at(m_player).end();
+  }
+};
+
+class BehaviorSupportProfile::Contingencies {
```
tturocy: We need to call this `BehaviorSupportProfile::SequenceContingencies` because there would also be a concept of `BehaviorContingencies` which could be iterated over, which is something different.

tturocy (approved):

Issue #612, opened 2025-11-14
BUG: Remove implementation of "dominated" actions


tturocy:
The concept of what constitutes a "dominated" action in an extensive game is rather subtle.  What is currently implemented is actually a mixture of a kind of "conditional" dominance and agent-form dominance (assuming subsequent actions are chosen independently as if in the agent-form).  This is neither here nor there.

As this is not a matter of simply correcting the implementation (but requiring an entirely new approach), we will remove this implementation.  It is actually only available at the moment via the GUI and was never exposed in Python.

So we will:
* Remove this from the GUI.  This includes also being able to remove the dominance elimination stack (which would need re-factoring anyway), and the layout features which support the display of partial trees (likewise something which would have needed attention independently).
* Remove this from the BehaviorSupportProfile, including removing any functions which exist solely to support this.

In a future version we would look to implement this type of analysis more cleanly - also possibly using sequences rather than actions (should there be results in that direction).

PR #620, opened 2025-11-17
Remove "dominated actions" method

tturocy:
See #612.

Done so far:

* Remove toolbar from extensive game window in GUI
* Remove computation of "dominated" actions in behavior support profiles, and the associated supporting code that was used only for this.

Still to do:

* [x] Review the documentation for the GUI to remove discussion of dominated actions.  This is slightly nontrivial because that discussion comes before that for strategies, so it requires editing not just deletion.  Also, the example for dominated strategies is not very interesting (prisoner's dilemma); it would be good to have a larger game which is less trivial (something dominance solvable in a few steps)?   A challenge here is the screenshots are fully 15 years old so making a new example would open that can of worms....
* [x] Remove support for "root reachability" in tree layout.  This will simplify the layout implementation substantially.  It is again something we might someday like to have, but it would be better to re-implement it in a better way from scratch.

### DIFF
--- a/src/gui/gameframe.cc
+++ b/src/gui/gameframe.cc
@@ -504,6 +504,9 @@ void GameFrame::MakeMenus()
   auto *toolsMenu = new wxMenu;
   toolsMenu->Append(GBT_MENU_TOOLS_DOMINANCE, _("&Dominance"), _("Find undominated actions"),
                     true);
+  if (m_doc->GetGame()->IsTree()) {
+    toolsMenu->Enable(GBT_MENU_TOOLS_DOMINANCE, false);
+  }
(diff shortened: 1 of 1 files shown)


rahulsavani:
> The issue of the menu did occur to me. What would you suggest for the behaviour? 

1. I think it is nice for it to be disabled when one cannot see the effect of changing it, i.e., whenever one is on the extensive view.
2. Having it checked but disabled in the extensive view can sensibly reflect that when one goes to the strategic view, the dominance stuff will be shown. By 1. though, we then cannot turn it on or off in the extensive view.

So then:

> It should be disabled menu item when viewing the tree? Should it retain its checkbox though? 

I think it should retain the check box, but it can't be changed from the extensive view.

> What I mean is: suppose you pass to strategic form, show the dominance toolbar, then go back to extensive. Would Dominance be disabled but still checked? 

Yes.

PR #621, opened 2025-11-17
Update tutorial 03 to replicate "Stripped down poker" example

edwardchalstrey1:
Includes:
- @rahulsavani final commits from `de-duplicate_simple_poker_game_in_test_suite` branch (see discussion in https://github.com/gambitproject/gambit/pull/608#issuecomment-3541821298)
- Updated tutorial 03 

Currently, the EFGs almost match, but with a couple of differences:

1. Tutorial version does not label nodes. @rahulsavani @tturocy I could add the node labels, however this raises the question of whether we'd rather demonstrate indexing with action labels (currently shown in tutorial) or via node labels, I thought based on previous discussions omitting node labels was preferred, though we could include both and explain in the tutorial that either can be used?
2. As discussed, terminal nodes reference player who won - (ok, now I'm thinking it's better to be consistent and label all nodes)

`stripped_down_poker.efg`:

```
EFG 2 R "A simple Poker game" { "Alice" "Bob" }
"Stripped-Down Poker: a simple game of one-card poker from Reiley et al (2008)."

c "" 1 "Deal" { "King" 1/2 "Queen" 1/2 } 0
p "" 1 1 "Alice has King" { "Bet" "Fold" } 0
p "" 2 1 "Bob's response" { "Call" "Fold" } 0
t "" 1 "Win Big" { 2, -2 }
t "" 2 "Win" { 1, -1 }
t "" 4 "Lose" { -1, 1 }
p "" 1 2 "Alice has Queen" { "Bet" "Fold" } 0
p "" 2 1 "Bob's response" { "Call" "Fold" } 0
t "" 3 "Lose Big" { -2, 2 }
t "" 2 "Win" { 1, -1 }
t "" 4 "Lose" { -1, 1 }
```

Tutorial EFG when saved:

```
EFG 2 R "Stripped-Down Poker: a simple game of one-card poker from Reiley et al (2008)." { "Alice" "Bob" }
""

c "" 1 "" { "King" 1/2 "Queen" 1/2 } 0
p "" 1 1 "" { "Bet" "Fold" } 0
p "" 2 1 "" { "Call" "Fold" } 0
t "" 1 "Alice wins big" { 2, -2 }
t "" 2 "Alice wins" { 1, -1 }
t "" 4 "Bob wins" { -1, 1 }
p "" 1 2 "" { "Bet" "Fold" } 0
p "" 2 1 "" { "Call" "Fold" } 0
t "" 3 "Bob wins big" { -2, 2 }
t "" 2 "Alice wins" { 1, -1 }
t "" 4 "Bob wins" { -1, 1 }
```

### DIFF
--- a/doc/tutorials/03_stripped_down_poker.ipynb
+++ b/doc/tutorials/03_stripped_down_poker.ipynb
@@ -189,7 +189,7 @@
    "metadata": {},
    "source": [
     "In game theory terms, this creates \"imperfect information\".\n",
-    "Bob cannot distinguish between these two nodes in the game tree, so he must use the same strategy (same probabilities for Call vs. Fold) in both situations.\n",
+    "Bob cannot distinguish between these two nodes in the game tree, so he must use the same same probabilities for Call vs. Fold in both situations.\n",
     "\n",
     "This is crucial in games where players must make decisions without full knowledge of the state of the game.\n",
     "\n",


tturocy:
Those terminal nodes don't have node labels, they have outcome labels.  Outcome labels make a lot of sense.

Information set labels sometimes make sense, because they are formally equivalence classes of objects (nodes).  Sometimes it's handy to stick a tag on those.

For nodes, a node is exactly equivalent to its history of actions (in a tree at least).  So the value of labelling nodes is minimal in the usual case.  It is better to think in terms of nodes as histories.

edwardchalstrey1:
> Those terminal nodes don't have node labels, they have outcome labels. Outcome labels make a lot of sense.
> 
> Information set labels sometimes make sense, because they are formally equivalence classes of objects (nodes). Sometimes it's handy to stick a tag on those.
> 
> For nodes, a node is exactly equivalent to its history of actions (in a tree at least). So the value of labelling nodes is minimal in the usual case. It is better to think in terms of nodes as histories.

Thanks @tturocy based on your comment then, I think the two EFGs in my comment above are close enough, marked this PR as ready for review

doc/tutorials/03_poker.ipynb
```diff
@@ -22,20 +22,20 @@
     "    - A card is dealt at random to Alice\n",
     "    - Alice observes her card\n",
```
rahulsavani: Missing full stops for the first sub-enumeration.

doc/tutorials/03_poker.ipynb
```diff
@@ -189,19 +189,19 @@
    "id": "689ce12c",
    "metadata": {},
    "source": [
-    "Now let's add the information set we created at the node where Alice raised with a King, to the node where Alice raised with a Queen."
```
rahulsavani, line 192: Betd -> Bet

doc/tutorials/03_poker.ipynb
```diff
@@ -160,27 +160,27 @@
...
+    "  - Chance player chooses King, then Alice Bets: `g.root.children[\"King\"].children[\"Bet\"]`\n",
+    "  - Chance player chooses Queen, then Alice Bets: `g.root.children[\"Queen\"].children[\"Bet\"]`\n",
     "\n",
-    "In other words, Bob's decision when Alice raises with a Queen should be part of the same information set as Bob's decision when Alice raises with a King.\n",
+    "In other words, Bob's decision when Alice Bets with a Queen should be part of the same information set as Bob's decision when Alice Bets with a King.\n",
     "\n",
     "To set this scenario up in Gambit, we'll need to use `Game.append_infoset` to add a move as part of an existing information set (represented in Gambit as an `Infoset`).\n",
     "\n",
-    "First, let's add Bob's move to the node where Alice has raised with a King."
+    "First, let's add Bob's move to the node where Alice has Betd with a King."
```
rahulsavani: We don't need append_infoset as we can just add the two nodes:

- g.root.children["King"].children["Bet"]
- g.root.children["Queen"].children["Bet"]

as a length-2 list to append_move and they will automatically be put in the same infoset.

rahulsavani (changes requested):
Looks good. A few minor comments:

- [x] Let's maybe call the file: 03_stripped_down_poker.ipynb (we will have other more complicated poker variants in the userguide/tutorials soon)

- [x] We cite as "[Mye91]" in the very first sentence. Let's cite Reiley et al (2008), which we already do just after the "enumeration" of subsections. That sentence could just move up to become the new first sentence. The, we can either drop all reference to Myerson, or have a footnote-like comment that the game is similar to but different from the version from Myserson'91.

- [x] Shall we not just unify the outcome labels of this tutorial and the game in the tests now? There are good reasons for either setup: in the tests everything is in terms of Alice, which is common in zero-sum games; in the tutorials, we use player names explicitly. I suggest to pick one, either, and just use it in both places.

- [x] Many cells have a colon in the sentence immediately preceding it, but many do not. I like having the colon where it makes sense, so perhaps try and be more consistent and add in more of these colons wherever possible.

- The final section "Representation of numerical data of a game" is really not about this poker example anymore, though it is clearly related. Fine to keep it but it could also be separate (and possibly then expanded into a full notebook). Might as well keep it as is for now though.

doc/tutorials/03_stripped_down_poker.ipynb
```diff
@@ -160,14 +160,13 @@
...
-    "\n",
-    "In other words, Bob's decision when Alice raises with a Queen should be part of the same information set as Bob's decision when Alice raises with a King.\n",
+    "  - Chance player chooses King, then Alice Bets: `g.root.children[\"King\"].children[\"Bet\"]`\n",
+    "  - Chance player chooses Queen, then Alice Bets: `g.root.children[\"Queen\"].children[\"Bet\"]`\n",
     "\n",
-    "To set this scenario up in Gambit, we'll need to use `Game.append_infoset` to add a move as part of an existing information set (represented in Gambit as an `Infoset`).\n",
+    "In other words, Bob's decision when Alice Bets with a Queen should be part of the same information set as Bob's decision when Alice Bets with a King.\n",
     "\n",
-    "First, let's add Bob's move to the node where Alice has raised with a King."
+    "To set this scenario up in Gambit, we'll need to use add both possible moves as part of the same information set (represented in Gambit as an `Infoset`).\n",
```
rahulsavani: we'll need to use add
->
we'll need to add

doc/tutorials/03_stripped_down_poker.ipynb
```diff
@@ -211,54 +189,54 @@
    "metadata": {},
    "source": [
     "In game theory terms, this creates \"imperfect information\".\n",
-    "Bob cannot distinguish between these two nodes in the game tree, so he must use the same strategy (same probabilities for Meet vs. Pass) in both situations.\n",
+    "Bob cannot distinguish between these two nodes in the game tree, so he must use the same strategy (same probabilities for Call vs. Fold) in both situations.\n",
```
rahulsavani: Technically a strategy is an object that specifies actions everywhere for a player. Better to just say:

so he must use the same probabilities for Call vs. Fold in both situations

tturocy:
Yes, what @edwardchalstrey1 said :)

rahulsavani:
Oh I see, the default, "squash and merge", is to squash everything to a single commit. 

And if one wants to do some squashing but still have multiple commits one can do an "interactive rebase". 

Sorry for noise, I only ever did this remotely with rebase.

PR #632, opened 2025-11-20
ENH: Implement pure strategy node reachability check

### LINKED ISSUE #629: ENH: Implement pure strategy node reachability check
**Description.**

The goal is to introduce a new method, `GameNode::IsStrategyReachable()`, to determine if a given node in an extensive tree form can be reached by any pure strategy profile. A node is reachable if there exists at least one pure strategy profile where the resulting path of play passes through that node. A pure strategy requires the player to choose the same action at every member node of that information set. In games with absent-mindedness, a path of play may visit the same information set multiple times, making some of the ancestors of its member nodes unreachable.

The existing `GameTreeRep::BuildInfosetParents()` method already handles the logic for absent-minded revisiting information sets. This traversal will be leveraged to collect the set of reachable decision nodes.

**Implementation Plan.**

1) In `GameTreeRep`, add member: 
`mutable std::set<GameNodeRep *> m_reachableNodes;`
2) Define the public API on the node, consistent with existing properties, e.g., `IsTerminal()` and `IsSubgameRoot()`:
In `GameNodeRep`, add the method bool `IsStrategyReachable() const;`
3) Modify `GameTreeRep::BuildInfosetParents()` to populate the `m_reachableNodes set`. 
4) Implement `GameNodeRep::IsStrategyReachable()` to trigger computation if needed and perform the check. 
5) Python API: Add a read-only property `Node.is_strategy_reachable`.

d-kad:
**Description.**

This pull request introduces a new method, `GameNode::IsStrategyReachable()`, to determine if a given node in an extensive tree form can be reached by any pure strategy profile. A node is reachable if there exists at least one pure strategy profile where the resulting path of play passes through that node. A pure strategy requires the player to choose the same action at every member node of that information set. In games with absent-mindedness, a path of play may visit the same information set multiple times, making some of the ancestors of its member nodes unreachable.

The existing `GameTreeRep::BuildInfosetParents()` method already handles the logic for absent-minded revisiting information sets. This traversal will be leveraged to collect the set of reachable decision nodes.

**Implementation Summary.**
The implementation follows the design discussed in the issue:

1) A new cache member `m_reachableNodes` is added to `GameTreeRep`
2) The existing `GameTreeRep::BuildInfosetParents()` traversal is augmented to populate this set, reusing its logic for handling absent-mindedness.
3) A new C++ method, `GameNodeRep::IsStrategyReachable()`, queries the cache
4) The functionality is exposed to Python as a new read-only property, `pygambit.Node.is_strategy_reachable`

Closes #629

### DIFF
--- a/src/games/gametree.cc
+++ b/src/games/gametree.cc
@@ -392,13 +392,12 @@ bool GameNodeRep::IsStrategyReachable() const
 {
   auto tree_game = static_cast<GameTreeRep *>(m_game);
 
-  if (tree_game->m_unreachableNodes.empty() && !tree_game->GetRoot()->IsTerminal()) {
+  if (!tree_game->m_unreachableNodes) {
     tree_game->BuildInfosetParents();
   }
 
   // A node is reachable if it is NOT in the set of unreachable nodes.
-  return tree_game->m_unreachableNodes.find(const_cast<GameNodeRep *>(this)) ==
-         tree_game->m_unreachableNodes.end();
+  return !contains(*tree_game->m_unreachableNodes, const_cast<GameNodeRep *>(this));
 }
 
 void GameTreeRep::DeleteParent(GameNode p_node)
@@ -813,7 +812,7 @@ void GameTreeRep::ClearComputedValues() const
   }
   const_cast<GameTreeRep *>(this)->m_nodePlays.clear();
   const_cast<GameTreeRep *>(this)->m_infosetParents.clear();
-  const_cast<GameTreeRep *>(this)->m_unreachableNodes.clear();
+  const_cast<GameTreeRep *>(this)->m_unreachableNodes = nullptr;
   m_computedValues = false;
 }
 
@@ -857,7 +856,7 @@ std::vector<GameNodeRep *> GameTreeRep::BuildConsistentPlaysRecursiveImpl(GameNo
 void GameTreeRep::BuildInfosetParents()
 {
   m_infosetParents.clear();
-  m_unreachableNodes.clear();
+  m_unreachableNodes = std::make_unique<std::set<GameNodeRep *>>();
 
   if (m_root->IsTerminal()) {
     m_infosetParents[m_root->m_infoset].insert(nullptr);
@@ -929,7 +928,7 @@ void GameTreeRep::BuildInfosetParents()
             while (!nodes_to_visit.empty()) {
               GameNodeRep *current_unreachable_node = nodes_to_visit.top();
               nodes_to_visit.pop();
-              m_unreachableNodes.insert(current_unreachable_node);
+              m_unreachableNodes->insert(current_unreachable_node);
 
               for (const auto &unreachable_child : current_unreachable_node->GetChildren()) {
                 nodes_to_visit.push(unreachable_child.get());
--- a/src/games/gametree.h
+++ b/src/games/gametree.h
@@ -40,7 +40,7 @@ protected:
   std::size_t m_numNonterminalNodes = 0;
   std::map<GameNodeRep *, std::vector<GameNodeRep *>> m_nodePlays;
   std::map<GameInfosetRep *, std::set<GameActionRep *>> m_infosetParents;
-  mutable std::set<GameNodeRep *> m_unreachableNodes;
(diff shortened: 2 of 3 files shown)


src/games/gametree.h
```diff
@@ -40,6 +40,7 @@ class GameTreeRep : public GameExplicitRep {
   std::size_t m_numNonterminalNodes = 0;
   std::map<GameNodeRep *, std::vector<GameNodeRep *>> m_nodePlays;
   std::map<GameInfosetRep *, std::set<GameActionRep *>> m_infosetParents;
+  mutable std::set<GameNodeRep *> m_reachableNodes;
```
tturocy: This is not a good way to implement this.  For almost every game this amounts to creating a complete copy of the tree.  Unreachable nodes are by far the exceptional case and not the rule, so this is too space-inefficient.

tturocy (changes requested):
Internally this really needs to be implemented as a set of nodes (or subtrees) which are not reachable.  Un-reachability is the exceptional case not the usual case.

tests/test_node.py
```diff
@@ -112,6 +112,29 @@ def test_legacy_is_subgame_root_set(game: gbt.Game, expected_result: set):
     assert legacy_roots == expected_roots
 
 
+@pytest.mark.parametrize("game, expected_reachable_indices", [
```
tturocy: I would recommend that we write the tests in terms of paths of action labels rather than indices into the list of nodes for two (related) reasons:

1.  if for some reason the iteration order of `nodes` were to change these tests would break.  While I don't anticipate we'd ever change the order in which we iterate over nodes it could happen.
2.  It is simply more transparent.

The best way to write the tests for (un)reachability of specific nodes would be to give the path as a list/tuple of action labels, and then iterate over that calling `.children` with the new action-label syntax (possibly writing a little helper function to do that) - eventually when we have out planned expanded syntax for sets of nodes it'll be possible to select nodes like that more elegantly.

tests/test_node.py
```diff
@@ -112,6 +112,29 @@ def test_legacy_is_subgame_root_set(game: gbt.Game, expected_result: set):
...
+    # Games without Absent-Mindedness where all nodes are reachable.
+    (games.read_from_file("e02.efg"), set(range(7))),
```
tturocy: A further thought on these tests.  By definition all nodes are reachable in a game that does not have absent mindedness.  This could be written as a different test that automatically just checks all nodes are reachable and that none of the information sets are absent-minded (once that feature is done).

tests/test_node.py
```diff
@@ -112,6 +112,57 @@ def test_legacy_is_subgame_root_set(game: gbt.Game, expected_result: set):
     assert legacy_roots == expected_roots
 
 
+def get_path_of_action_labels(node: gbt.Node) -> list[str]:
```
tturocy: Should probably preface this function name with an underscore as it's "private".

src/games/gametree.cc
```diff
@@ -388,6 +388,19 @@ bool GameNodeRep::IsSubgameRoot() const
   return true;
 }
 
+bool GameNodeRep::IsStrategyReachable() const
+{
+  auto tree_game = static_cast<GameTreeRep *>(m_game);
+
+  if (tree_game->m_unreachableNodes.empty() && !tree_game->GetRoot()->IsTerminal()) {
```
tturocy: If I'm reading this correctly, in a game with no absent-mindedness this will ALWAYS try to re-build the unreachable nodes on every call, because `m_unreachableNodes` will always be empty?

The trick for doing such caching would be to store the cached value in a `std::unique_ptr`.  Then you can test whether the cached pointer is null or not, and to invalidate the cache you just reset the pointer and deallocation is automatic.

src/games/gametree.cc
```diff
@@ -388,6 +388,19 @@ bool GameNodeRep::IsSubgameRoot() const
...
+    tree_game->BuildInfosetParents();
+  }
+
+  // A node is reachable if it is NOT in the set of unreachable nodes.
```
tturocy, line 399: We have a utility function `contains` (defined in `core/util.h`) which expresses this more idiomatically while we await future versions of C++ that will implement it natively.

tturocy (changes requested):

PR #634, opened 2025-11-21
Fix floating formatting

### LINKED ISSUE #633: [Bug]: Rational cannot be formatted using float specification in notebook tests

PyGambit

macOS

Version: 16.4.0

### What happened?

When running `pytest tests/test_tutorials.py` I get the following error:

```
E           ---------------------------------------------------------------------------
E           TypeError                                 Traceback (most recent call last)
E           Cell In[30], line 7
E                 2 print(
E                 3     f"{player.label}'s expected payoffs:"
E                 4 )
E                 5 for action in player.actions:
E                 6     print(
E           ----> 7         f"At information set {action.infoset.number}, "
E                 8         f"when playing {action.label} - "
E                 9         f"gnm: {gnm_eqm.as_behavior().action_value(action):.4f}"
E                10         f", lcp: {eqm.action_value(action):.4f}"
E                11     )
E                12 print()
E           
E           TypeError: unsupported format string passed to Rational.__format__
```

d-kad:
Convert Rational to float before using string formatting

Closes #633

### DIFF
--- a/doc/tutorials/03_poker.ipynb
+++ b/doc/tutorials/03_poker.ipynb
@@ -963,8 +963,8 @@
     "        print(\n",
     "            f\"At information set {action.infoset.number}, \"\n",
     "            f\"when playing {action.label} - \"\n",
-    "            f\"gnm: {float(gnm_eqm.as_behavior().action_value(action)):.4f}\"\n",
-    "            f\", lcp: {float(eqm.action_value(action)):.4f}\"\n",
+    "            f\"gnm: {gnm_eqm.as_behavior().action_value(action):.4f}\"\n",
+    "            f\", lcp: {eqm.action_value(action)}\"\n",
     "        )\n",
     "    print()"
    ]


tturocy:
I do not understand why this is needed?  `gnm_solve` returns a `MixedStrategyProfileDouble` which already expresses everything as `float`.

When I run the tutorial myself it works fine.

tturocy:
I wouldn't think this would have to do with Python version.  Those numbers are always going to be `float`.  My feeling is it is more possible that it is some kind of out-of-order execution that is causing that variable not to be the result of `gnm_solve`?

But also I don't think the solution is the right one anyway.  If in fact we have a situation where the data are rational we should print them as rational, not cast them to `float`.  The whole point in computing with rationals is for precision!  So if we did have such a situation the remedy I'd recommend is not to use the float-format specifier in the f-string.  Rationals do have a `__str__` and a `__repr__` that should do the right thing in a f-string.

d-kad:
> > I do not understand why this is needed? `gnm_solve` returns a `MixedStrategyProfileDouble` which already expresses everything as `float`.
> > When I run the tutorial myself it works fine.
> 
> Same here actually which is why this didn't come up before, I wonder if this is also related to python version somehow? @d-kad which python version are you running locally?

Python 3.9.6

tturocy:
The thing is `gnm_eqm` ought to be a double-precision profile no matter what, so I'm not seeing why it would be generating a rational here.

But also @d-kad see #635 - 3.9 is now EOL and we won't be supporting it for this release, time to upgrade! :)

Issue #637, opened 2025-11-21
MAINT: Remove C++ solve-by-subgames


tturocy:
We are implementing a new approach to identifying proper subgames (see #567).  As part of that, there will be a more natural way of working with upgames.

Solving a game by subgames is a process that ideally should be controlled more by the user, as there are choices to be made.  For example:

1.  The user might want to use different solution methods in different subgames.  An "upgame" or "subgame difference" with just one player is of course trivial to solve and it would be overkill blindly to apply a solution method.  Likewise, there are advantageous algorithms for two-player subgame differences that could be used even if the larger game has more than two players.
2.  The user might want to filter some subgame equilibria/outcomes.
3.  When proper subgames have multiple equiilbria, the process naturally lends itself to parallelisation.

The existing C++ implementation permits basically none of this, and is awkwardly implemented to boot.

Also, it is hardly exposed (only in the command-line tools of `enumpure`, `lcp` and `lp`.

Therefore, we're going to remove this entirely from C++.  We will instead replace this with some sample codes in Python which implement the procedure to develop the idea further.

PR #641, opened 2025-11-24
Remove C++ subgame-perfect equilibrium solver

tturocy:
This implements #637.  It also includes some minor refactoring of the command-line tool drivers which it was convenient to do while making this change in those files.

### DIFF
--- a/src/tools/convert/convert.cc
+++ b/src/tools/convert/convert.cc
@@ -22,11 +22,12 @@
 
 #include <iostream>
 #include <fstream>
-#include <cstdlib>
 #include <getopt.h>
 
 #include "gambit.h"
 #include "games/writer.h"
+
+using namespace Gambit;
 
 void PrintBanner(std::ostream &p_stream)
 {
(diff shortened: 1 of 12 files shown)


tturocy:
@rahulsavani for you to check please that everything seems to work for you with the affected tools.

tturocy:
> I tested gambit-enumpure, gambit-lp, and gambit-lcp on master and on this branch.
> 
> I can confirm that the -P option is now gone and all three seem to work fine now, so this PR looks good.
> 
> I will note however, that with all three CLI tools, I could not get them to work with the -P option on current master. I would always get a "Error: Dereferenced null pointer", as in:
> 
> ```
> gambit-enumpure -P sample_extensive_game.efg
> Search for Nash equilibria in pure strategies
> Gambit version 16.4.0, Copyright (C) 1994-2025, The Gambit Project
> This is free software, distributed under the GNU GPL
> 
> Error: Dereferenced null pointer
> ```
> 
> Not our problem any more though once this is merged in I suppose ;-)

But it is!  Now see #647.

PR #656, opened 2025-12-02
Fix dependency installation in python.yml GitHub Action

edwardchalstrey1:
### Changes

- Now GH action installs from the `requirements.txt` files used by tests and docs directly
- Also moves the logic that skips running the tutorial tests from the GH action to the test file itself (don't run for <3.12), to avoid any inconsistencies with GH actions PR vs Push vs running locally
- Removes the version pinning for doc and test requirements - this was actually preventing the installation of packages from working across Python versions, as the pinned package versions for earlier python versions were not always available. It was too stringent to pin these and I don't think there's much to be gained from having done so - if the latest versions of any of those packages do end up including changes that break our docs infrastructure, this will at worst mean that a docs build could fail, but that would be picked up in any PR and would not affect the current build. Might be worth discussing pinning of dependencies of `pygambit` itself in future though
- See comment below: The `paths-ignore` in the GitHub action yml was wrongly preventing some checks from running. I don't think it's a big problem to remove this completely, whilst PRs take a long time to run, we don't do docs PRs in high enough volume/speed for this to be too much of a problem
- Since Windows requires a specific manual installation of OpenSpiel, I have added a Warning that links to this in the [How to run PyGambit tutorials on your computer](https://gambitproject--656.org.readthedocs.build/en/656/tutorials/running_locally.html#local-tutorials) and skipped running OpenSpiel notebook when running tests on Windows GH action. Also made a few other edits to this page whilst I was there.

### DIFF
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -102,7 +102,7 @@ jobs:
           python -m pip install --upgrade pip
           pip install setuptools build cython wheel
           pip install -r tests/requirements.txt
-          pip install -r doc/requirements.txt
+          pip install nbformat nbclient ipykernel jupyter matplotlib
       - name: Build extension
         run: |
           python -m pip install -v .
--- a/tests/test_tutorials.py
+++ b/tests/test_tutorials.py
@@ -50,6 +50,11 @@ def test_execute_notebook(nb_path):
     if sys.version_info < (3, 12):
         pytest.skip("Notebook execution tests require Python 3.12 or newer")
 
+    # Skip OpenSpiel notebook on Windows
+    # (OpenSpiel is not available on Windows without manual install)
+    if sys.platform == "win32" and "openspiel" in nb_path.name.lower():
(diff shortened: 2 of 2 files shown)


edwardchalstrey1:
Hmm it seems that the rules I added for when to run python checks are incorrectly preventing those checks from running, even when we do actually edit files outside the excluded folders in some cases... hence problems were missed when merging previous PRs:

```yml
on:
  push:
    paths-ignore:
      - 'doc/**'
      - '.github/ISSUE_TEMPLATE/**'
      - 'README.md'
  pull_request:
    paths-ignore:
      - 'doc/**'
      - '.github/ISSUE_TEMPLATE/**'
      - 'README.md'
 ```

For now I'm going to remove these `paths-ignore` checks completely and address the address the issue @tturocy noticed in #655 and ensure matplotlib is included in the requirements

edwardchalstrey1:
Ah, another issue now the checks are running properly I didn't notice before, OpenSpiel install is throwing an error on Windows!

PR #664, opened 2025-12-04
 Implement efficient single-pass subgame root finder to replace the legacy `IsSubgameRoot`

### LINKED ISSUE #584: FEAT: Implement improved algorithm for finding subgame roots

_No response_

We now have developed a new algorithm that finds all subgame roots with one pass of the tree.  This algorithm is essentially as efficient as what we currently do to determine if a **given** node is a subgame root.

Implement this new algorithm (cacheing the results as appropriate), and use it to replace the current implementation of `GameNodeRep::IsSubgameRoot` as appropriate.  (Given the `friend` structure in C++ it may make sense internally to move this from `GameNodeRep` to `GameTreeRep`.)

d-kad:
**Description**

This PR replaces the legacy implementation of `IsSubgameRoot` with a new single-pass algorithm.
Computes and caches subgame roots for all infosets in one traversal.

**Key Changes**

- Implemented `FindSubgameRoots` using bottom-up traversal (DSU + Priority Queue)
- Added  DSU helper functions in an anonymous namespace
- Added `GenerateComponent` and `SubgameScratchData` helpers in anonymous namespace.
- Added `BuildSubgameRoots` to populate and cache `m_infosetSubgameRoot`
- Refactored existing subgame tests to use action-path comparisons and added new test games

Closes #584

### DIFF
--- a/src/games/gametree.cc
+++ b/src/games/gametree.cc
@@ -1157,16 +1157,19 @@ struct SubgameScratchData {
 ///
 /// @param p_data The DSU data structure to update with component information
 /// @param p_start_node The node to start component generation from
+/// @param p_node_ordering A map providing a strict total ordering (DFS Preorder) of nodes
 ///
 /// Precondition: p_start_node must be non-terminal
 /// Precondition: p_start_node's infoset must not already be in p_data.dsu_parent
 /// Postcondition: p_data.subgame_root_candidate[leader] contains the highest node
 ///                in the newly-formed component
-void GenerateComponent(SubgameScratchData &p_data, GameNodeRep *p_start_node)
+void GenerateComponent(SubgameScratchData &p_data, GameNodeRep *p_start_node,
+                       const std::unordered_map<const GameNodeRep *, int> &p_node_ordering)
 {
-  auto node_cmp = [](const GameNodeRep *a, const GameNodeRep *b) {
-    return a->GetNumber() < b->GetNumber();
+  auto node_cmp = [&p_node_ordering](const GameNodeRep *a, const GameNodeRep *b) {
+    return p_node_ordering.at(a) < p_node_ordering.at(b);
   };
+
   std::priority_queue<GameNodeRep *, std::vector<GameNodeRep *>, decltype(node_cmp)>
       local_frontier(node_cmp);
 
@@ -1234,14 +1237,24 @@ std::map<GameInfosetRep *, GameNodeRep *> FindSubgameRoots(const Game &p_game)
     return {};
   }
 
+  // Pre-compute DFS numbers locally.
+  std::unordered_map<const GameNodeRep *, int> node_ordering;
+  int counter = 0;
+  for (const auto &node : p_game->GetNodes(TraversalOrder::Preorder)) {
+    node_ordering[node.get()] = counter++;
+  }
+
   SubgameScratchData data;
 
+  // Define filter predicate
+  auto is_unvisited_infoset = [&data](const auto &node) {
+    return !data.dsu_parent.count(node->GetInfoset().get());
+  };
+
   // Process nodes in postorder
-  for (const auto &node : filter_if(p_game->GetNonterminalNodes(TraversalOrder::Postorder),
-                                    [&data](const auto &node) {
-                                      return !data.dsu_parent.count(node->GetInfoset().get());
-                                    })) {
-    GenerateComponent(data, node.get());
+  for (const auto &node :
+       filter_if(p_game->GetNonterminalNodes(TraversalOrder::Postorder), is_unvisited_infoset)) {
+    GenerateComponent(data, node.get(), node_ordering);
   }
 
   std::map<GameInfosetRep *, GameNodeRep *> result;


src/games/gametree.cc
```diff
@@ -1068,6 +1069,216 @@ void GameTreeRep::BuildUnreachableNodes()
...
+  std::map<GameInfosetRep *, int> infoset_to_id;
+
+  // Visited tracking
+  std::vector<bool> infoset_visited;
+  std::vector<int> node_visited_token;
+  int current_token = 0;
+};
+
+// DSU Find with path compression
+int FindSet(SubgameScratchData &p_data, int p_i)
```
tturocy: Seems like this is better written as a member of the scratch data structure, as it's an operation on the structure itself.

Does this need to be recursive?  Recursion is much more expensive than simple iteration.

src/games/gametree.cc
```diff
@@ -1068,6 +1069,216 @@ void GameTreeRep::BuildUnreachableNodes()
...
+{
+  if (p_data.dsu_parent[p_i] != p_i) {
+    p_data.dsu_parent[p_i] = FindSet(p_data, p_data.dsu_parent[p_i]);
+  }
+  return p_data.dsu_parent[p_i];
+}
+
+// DSU Union - merge set j into set i, tracking best subgame root
+void UnionSets(SubgameScratchData &p_data, int p_i, int p_j, GameNodeRep *p_current_node)
```
tturocy: Likewise seems like a member operation

src/games/gametree.cc
```diff
@@ -1068,6 +1069,216 @@ void GameTreeRep::BuildUnreachableNodes()
...
+};
+
+// Scratch data structure for subgame root finding algorithm
+struct SubgameScratchData {
+  // DSU structures
+  std::vector<int> dsu_parent;
+  std::vector<GameNodeRep *> leader_subgame_map;
+
+  // ID mapping
+  std::map<GameInfosetRep *, int> infoset_to_id;
```
tturocy: We should be avoiding the use of integer IDs, and doubly so if we have to go through work to compute them.  Everything ought to be doable efficiently in terms of references to game objects.

Doing it this way means we will have to re-do it entirely for Gambit 17.  I'd rather us write it the best way first time.

src/games/gametree.cc
```diff
@@ -1068,6 +1069,216 @@ void GameTreeRep::BuildUnreachableNodes()
...
+      }
+    }
+  }
+
+  // Process components in descending node order
+  while (!global_frontier.empty()) {
+    auto *node = global_frontier.top();
+    global_frontier.pop();
+
+    const int id = data.infoset_to_id[node->GetInfoset().get()];
```
tturocy: Here's an example of where this is all more complicated than it needs to be, because in some places we are using game objects and others we are using arbitrary integers, instead of just using the game objects.

tests/test_node.py
```diff
@@ -213,11 +191,155 @@ def test_node_own_prior_action_non_terminal(game_file, expected_node_data):
...
+    # Games with Absent-Mindedness where the legacy method is known to fail.
+    pytest.param(
+        games.read_from_file("AM-subgames.efg"),
+        {0, 2, 5, 8},  # The correct set of roots
+        marks=pytest.mark.xfail(
+            reason="Legacy method does not detect subgame roots that are members of AM-infosets."
+        )
+    ),
+])
+def test_legacy_is_subgame_root_set(game: gbt.Game, expected_result: set):
```
tturocy: This shouldn't be called "legacy".  We are still going to have `is_subgame_root`.  Just the implementation is different.

src/games/gametree.h
```diff
@@ -47,6 +47,7 @@ class GameTreeRep : public GameExplicitRep {
   std::map<GameNodeRep *, std::vector<GameNodeRep *>> m_nodePlays;
   mutable std::shared_ptr<OwnPriorActionInfo> m_ownPriorActionInfo;
   mutable std::unique_ptr<std::set<GameNodeRep *>> m_unreachableNodes;
+  std::map<GameInfosetRep *, GameNodeRep *> m_infosetSubgameRoot;
```
tturocy: Doesn't this need to be mutable and the cache removed when the game is changed?

src/games/gametree.cc
```diff
@@ -1068,6 +1069,216 @@ void GameTreeRep::BuildUnreachableNodes()
   }
 }
 
+//------------------------------------------------------------------------
+//                        Subgame Root Finder
+//------------------------------------------------------------------------
+
+namespace { // Anonymous namespace for internal implementation
+
+// Comparator for priority queue (ancestors have lower priority than descendants)
```
tturocy: Just checking that this is OK.  Comparing numbers is not the same as the ancestor partial ordering.  It's true that if a is an ancestor of b then it has a lower number, but the reverse is not true.

src/games/game.h
```diff
@@ -432,6 +432,7 @@ class GameNodeRep : public std::enable_shared_from_this<GameNodeRep> {
   friend class GameInfosetRep;
   friend class GamePlayerRep;
   friend class PureBehaviorProfile;
+  friend class SubgameRootFinder;
```
tturocy: This class doesn't exist anymore, right?

src/games/game.h
```diff
@@ -759,6 +760,8 @@ class GameRep : public std::enable_shared_from_this<GameRep> {
 
   /// Returns true if the game is perfect recall
   virtual bool IsPerfectRecall() const = 0;
+  ///
+  virtual GameNode GetSubgameRoot(GameInfoset) const = 0;
```
tturocy: Slightly better to write `const GameInfoset &` technically to avoid a copy (not all functions have been improved to this convention but they will be)

src/games/gametree.cc
d-kad: Refactored FindSet to be iterative (can think of examples with a long chain of expanding components)

src/games/gametree.cc
d-kad: I refactored to use game object pointers instead of the integer IDs for infosets in DSU.
Node numbers only used for array indexing (intrinsic ``GetNumber()``, not computed)

src/games/game.h
```diff
@@ -759,6 +759,8 @@ class GameRep : public std::enable_shared_from_this<GameRep> {
 
   /// Returns true if the game is perfect recall
   virtual bool IsPerfectRecall() const = 0;
+  ///
+  virtual GameNode GetSubgameRoot(const GameInfoset &) const = 0;
```
tturocy: The preferred pattern is to raise `UndefinedException` in the base class; that way it is not necessary to modify every potential child class to override it, especially for representations where this operation does not make sense.

src/games/gametree.cc
```diff
@@ -371,32 +372,10 @@ bool GameNodeRep::IsSuccessorOf(GameNode p_node) const
 
 bool GameNodeRep::IsSubgameRoot() const
 {
-  // First take care of a couple easy cases
-  if (m_children.empty() || m_infoset->m_members.size() > 1) {
+  if (m_children.empty()) {
```
tturocy, line 376: Except we agreed that for the trivial game with no nodes the root node is in fact a subgame root!

src/games/gametree.cc
```diff
@@ -1078,6 +1058,182 @@ void GameTreeRep::BuildUnreachableNodes()
...
+  // DSU structures
+  std::unordered_map<GameInfosetRep *, GameInfosetRep *> dsu_parent;
+  std::unordered_map<GameInfosetRep *, GameNodeRep *> subgame_root_candidate;
+  // Timestamps for nodes
+  std::vector<int> node_visited_token;
+  int current_token = 0;
+
+  std::vector<GameNodeRep *> local_frontier;
+
+  // DSU Find with path compression
```
tturocy: Because this algorithm is a bit nontrivial, we need clear descriptions of the component functions, including ideally pre-conditions and post-conditions where appropriate.

src/games/gametree.cc
```diff
@@ -1078,6 +1058,182 @@ void GameTreeRep::BuildUnreachableNodes()
...
+  GameInfosetRep *FindSet(GameInfosetRep *p_infoset)
+  {
+    // Initialize/retrieve current parent
+    auto *leader = dsu_parent.try_emplace(p_infoset, p_infoset).first->second;
+    while (dsu_parent[leader] != leader) {
```
tturocy: Accesses to a map should always be using `.at` and not the `[]` operator.  The `[]` operator creates an entry if one does not exist - which can make some subtle bugs difficult to trace!

src/games/gametree.cc
```diff
@@ -1078,6 +1058,182 @@ void GameTreeRep::BuildUnreachableNodes()
...
+    frontier[node_num] = node;
+    p_data.node_visited_token[node_num] = p_data.current_token;
+    active_nodes_counter++;
+  };
+
+  push_node(p_start_node);
+
+  // Process nodes in descending order
+  for (int node_num = max_node_number; node_num >= min_node_number; --node_num) {
+    auto *curr = frontier[node_num];
```
tturocy: This just smells bad.  The frontier is still a vector that's the same size as the game.   Possibly in the worst case there's nothing we can do about that.  But in a game with proper subgames this is going to be carrying around a lot of unnecessary information and is going to scale really poorly.

src/games/gametree.cc
```diff
@@ -1078,6 +1058,182 @@ void GameTreeRep::BuildUnreachableNodes()
...
+    min_node_number = std::min(min_node_number, node_num);
+    frontier[node_num] = node;
+    p_data.node_visited_token[node_num] = p_data.current_token;
+    active_nodes_counter++;
+  };
+
+  push_node(p_start_node);
+
+  // Process nodes in descending order
+  for (int node_num = max_node_number; node_num >= min_node_number; --node_num) {
```
tturocy: In modern C++ we're almost never going to iterate like this.  This ought to be at least a reverse range iterator.  Which combined with changing the data structure for the frontier would also solve the smelliness of checking for null entries, because if this were e.g. just a map we would automatically iterate over only those with are actually on the frontier.

src/games/gametree.cc
```diff
@@ -1078,6 +1058,182 @@ void GameTreeRep::BuildUnreachableNodes()
...
+    auto *curr = frontier[node_num];
+    if (!curr) {
+      continue; // Skip empty slots
+    }
+
+    active_nodes_counter--;
```
tturocy: We really should not be manually counting this; this should be the size of some container no?

src/games/gametree.cc
```diff
@@ -1078,6 +1058,182 @@ void GameTreeRep::BuildUnreachableNodes()
...
+    p_data.UnionSets(start_infoset, curr_infoset, curr);
+
+    if (is_external_collision) {
+      auto *candidate_root = p_data.subgame_root_candidate[p_data.FindSet(curr_infoset)];
+      if (p_data.node_visited_token[candidate_root->GetNumber()] != p_data.current_token) {
+        push_node(candidate_root);
+      }
+    }
+    else {
+      for (const auto &member_sp : curr->GetInfoset()->GetMembers()) {
```
tturocy: Assuming this doesn't completely change with respect to other comments, this can be written as a filtered iterator combined with a suitable version of `std::transform` I think.

src/games/gametree.cc
```diff
@@ -1078,6 +1058,182 @@ void GameTreeRep::BuildUnreachableNodes()
...
+  const int num_nodes = p_game->NumNodes();
+
+  // Allocate local frontier once for reuse across all component generations
+  data.local_frontier.resize(num_nodes + 1, nullptr);
+  data.node_visited_token.resize(num_nodes + 1, 0);
+
+  // Collect nodes in ascending order (depth-first traversal)
+  std::vector<GameNodeRep *> nodes_ascending;
+  nodes_ascending.reserve(num_nodes);
+  for (const auto &node : p_game->GetNodes()) {
```
tturocy: Again this is copying the entire tree, which really isn't going to be acceptable for performance (similar to comments on the previous version).

src/games/gametree.cc
d-kad: Do we want to replace the combination of for/if statements with std::transform and a filtering iterator?
If the comment still stands for the new implementation, could you please give more details as the solutions I see are pretty bulky.

src/games/gametree.cc
d-kad: will be replaced with the new Nodes iterator -- for now left as is with a TODO comment

src/games/gametree.cc
```diff
@@ -1093,6 +1074,205 @@ void GameTreeRep::BuildUnreachableNodes() const
...
+std::map<GameInfosetRep *, GameNodeRep *> FindSubgameRoots(const Game &p_game)
+{
+  if (p_game->GetRoot()->IsTerminal()) {
+    return {};
+  }
+
+  SubgameScratchData data;
+
+  // Process nodes in postorder
+  for (const auto &node : p_game->GetNonterminalNodes(TraversalOrder::Postorder)) {
```
tturocy: This can use the new `filter_if` defined in `core/util.h` so we are C++20-ranges ready here.

src/games/gametree.cc
```diff
@@ -1093,6 +1074,205 @@ void GameTreeRep::BuildUnreachableNodes() const
...
+  }
+
+  return result;
+}
+
+} // end anonymous namespace
+
+void GameTreeRep::BuildSubgameRoots() const
+{
+  if (!m_computedValues) {
```
tturocy: Why do we need to call this?  As we were discussing in the catchup this week, `ComputedValues` actually just means "build reduced strategies".  We don't need reduced strategies for subgame roots.

src/games/gametree.cc
```diff
@@ -1093,6 +1074,205 @@ void GameTreeRep::BuildUnreachableNodes() const
...
+  for (const auto &node : p_game->GetNonterminalNodes(TraversalOrder::Postorder)) {
+    if (data.dsu_parent.count(node->GetInfoset().get())) {
+      continue;
+    }
+    GenerateComponent(data, node.get());
+  }
+
+  std::map<GameInfosetRep *, GameNodeRep *> result;
+
+  for (const auto &player : p_game->GetPlayersWithChance()) {
```
tturocy: This can be done by iterating over a `NestedElementCollection` - see definition of `Infosets` in `GameRep` and `gameobject.h`.

src/games/gametree.cc
```diff
@@ -1093,6 +1074,205 @@ void GameTreeRep::BuildUnreachableNodes() const
...
+//------------------------------------------------------------------------
+//                        Subgame Root Finder
+//------------------------------------------------------------------------
+
+namespace { // Anonymous namespace
+
+// Comparator for priority queue (ancestors have lower priority than descendants)
+// This ensures descendants are processed first, allowing upward traversal to find
+// the highest reachable node in each component
+struct NodeCmp {
```
tturocy: From a C++ perspective, can't we just get away with a lambda in the definition of a priority queue, without having to go through the palaver of a struct with call defined?

The problem with relying on numbers here is that nodes are only numbered in `SortInfosets`.  And you're not calling `SortInfosets`.  So your tests are only working by accident, because you are only working with games that have been loaded from a file (and those are automatically sorted).

We really need to replace this with something which does not require all nodes to be numbered at the start.  Otherwise we are making yet another pass through the tree to do this.

d-kad: @rahulsavani @tturocy In this version, I added a local pre-pass to number the nodes. This removed the dependency on `SortInfosets` (triggered by `BuildComputedValues`) while keeping numbers used by the algorithm.

For context on the comparator: I attempted to use node depth to avoid the pre-pass, but this resulted in 6 test failures. The order of nodes on the local frontier was arbitrary when they had the same depth, breaking the merging logic (more on this and lexicographic comparison of nodes' action paths in my comment below).

d-kad:
@tturocy @rahulsavani Would this be a convincing property?

**Technically**: We pop the node with the highest DFS preorder number from the local frontier.

**Structurally**: This means processing the deepest, leftmost node in the local frontier at each step, its lower-left corner.
Because of the further addition of nodes to the frontier (infoset members and parents), the corner is updated at each step.

**Formal characterization**: Consider paths of action numbers to nodes in the tree. Each time a node is popped from the local frontier, it has the lexicographically maximal path among all nodes on this frontier. The first dimension of comparison is depth; for two nodes with equal depth, the leftmost one is considered greater.

Why not use lexicographic path comparison directly?

**Drawbacks**:
* Such lexicographic ordering would rely on action numbers (Gambit's internal numbering), which we try to avoid as an implementation detail;
* Lexicographic comparison is more expensive than integer comparison (O(path length) vs O(1)).

Pre-computing DFS numbers may thus be an optimal choice: it trades one upfront tree traversal with constant comparisons for the same tree traversal with storing paths (more memory intensive than storing numbers) and expensive lexicographic comparisons.

tturocy:
This has been superseded by #750.

PR #676, opened 2025-12-08
ENH: Add is_absent_minded property to Infoset

d-kad:
**Description:**
Implement the absent-minded checker at the level of individual information sets.

**Changes:**
- Implemented ``GameTreeRep::IsAbsentMinded(const GameInfosetRep *infoset)`` to check if a given infoset has absent-mindedness. The method relies on ``BuildUnreachableNodes()``.
- Exposed the C++ method as a property ``is_absent_minded`` on the ``pygambit.Infoset class``.
- Refactored the absent-mindedness test suite to verify the property on specific Infoset objects rather than the Game

Closes: #578

### DIFF
--- a/src/pygambit/infoset.pxi
+++ b/src/pygambit/infoset.pxi
@@ -168,6 +168,8 @@ class Infoset:
 
         An information set is absent-minded if there exists a path of play
         in the game tree that intersects the information set more than once.
+
+        .. versionadded:: 16.5.0
         """
         return self.infoset.deref().GetGame().deref().IsAbsentMinded(self.infoset)
 


src/games/gametree.cc
```diff
@@ -207,6 +207,15 @@ bool GameInfosetRep::Precedes(GameNode p_node) const
   return false;
 }
 
+bool GameInfosetRep::IsAbsentMinded() const
```
tturocy: We have just been defining these properties on the `GameRep` directly rather than on the classes as part of our evolution towards version 17.  Doing so avoids the dance of having to do pointer casting just to forward it to the game class.

src/games/gametree.cc
```diff
@@ -804,6 +813,19 @@ Rational GameTreeRep::GetPlayerMaxPayoff(const GamePlayer &p_player) const
       p_player, m_root, [](const Rational &a, const Rational &b) { return std::max(a, b); });
 }
 
+// bool GameTreeRep::IsAbsentMinded() const
```
tturocy: We can get rid of this as agreed.

src/games/gametree.cc
```diff
@@ -819,6 +841,15 @@ bool GameTreeRep::IsPerfectRecall() const
                      [](const auto &pair) { return pair.second.size() <= 1; });
 }
 
+bool GameTreeRep::IsAbsentMinded(const GameInfosetRep *infoset) const
```
tturocy: This should be the public member of the game, taking a `const GameInfoset &p_infoset`.

src/games/gametree.h
```diff
@@ -47,6 +47,7 @@ class GameTreeRep : public GameExplicitRep {
   std::map<GameNodeRep *, std::vector<GameNodeRep *>> m_nodePlays;
   mutable std::shared_ptr<OwnPriorActionInfo> m_ownPriorActionInfo;
   mutable std::unique_ptr<std::set<GameNodeRep *>> m_unreachableNodes;
+  std::set<GameInfosetRep *> m_absentMindedInfosets;
```
tturocy: Should be `mutable` as it is a cached property, so we don't need to cast away const.

src/pygambit/infoset.pxi
```diff
@@ -161,6 +161,11 @@ class Infoset:
         """Whether the information set belongs to the chance player."""
         return self.infoset.deref().IsChanceInfoset()
 
+    @property
+    def is_absent_minded(self) -> bool:
+        """Whether the information set has absent-mindedness."""
+        return self.infoset.deref().IsAbsentMinded()
```
tturocy: With the changes as above this will become a call to the member of the game.

tests/test_infosets.py
```diff
@@ -162,3 +162,31 @@ def test_infoset_own_prior_actions(game_file, expected_results):
...
+    ("noPR-action-AM.efg", {("Player 1", 0)}),
+    ("noPR-action-AM-two-hops.efg", {("Player 1", 0), ("Player 2", 0)}),
+])
+def test_infoset_is_absent_minded(game_input, expected_infosets):
+    """
+    Verify the is_absent_minded property of information sets.
+    """
+    game = games.read_from_file(game_input)
+
+    actual_infosets = {
```
tturocy: I suggest a more transparent approach is to refer to an information set by one of its members.  With members we can use the path of action labels to that member (as we have done elsewhere) which is more intuitive.  So instead of relying on information set numbers we can say "the information set that is reached after a given sequence of action labels" which is going to be more robust.

src/games/game.h
```diff
@@ -759,6 +759,8 @@ class GameRep : public std::enable_shared_from_this<GameRep> {
 
   /// Returns true if the game is perfect recall
   virtual bool IsPerfectRecall() const = 0;
+  /// Returns true if an infoset of a given game is absent-minded
+  virtual bool IsAbsentMinded(const GameInfoset &) const { return false; }
```
tturocy: If we're going to return false as a default (which is sensible), we do need to check that the infoset belongs to this game.  Otherwise you could have an absent-minded infoset, but call this function on a different game, and get "false"

src/games/gametree.cc
```diff
@@ -870,6 +883,7 @@ void GameTreeRep::ClearComputedValues() const
   const_cast<GameTreeRep *>(this)->m_nodePlays.clear();
   m_ownPriorActionInfo = nullptr;
   const_cast<GameTreeRep *>(this)->m_unreachableNodes = nullptr;
+  const_cast<GameTreeRep *>(this)->m_absentMindedInfosets.clear();
```
tturocy: We don't need to cast away const here.

src/pygambit/infoset.pxi
```diff
@@ -161,6 +161,11 @@ class Infoset:
         """Whether the information set belongs to the chance player."""
         return self.infoset.deref().IsChanceInfoset()
 
+    @property
+    def is_absent_minded(self) -> bool:
+        """Whether the information set has absent-mindedness."""
```
tturocy: We should be consistent about terminology.  Is it "is + <adjective>" or "has + <noun>".

A slightly longer second part of the docstring would do well to explain what it means for an information set to be absent-minded.

tests/test_infosets.py
```diff
@@ -162,3 +162,44 @@ def test_infoset_own_prior_actions(game_file, expected_results):
         }
 
         assert actual_details == expected_set
+
+
+def _get_infoset_by_path(game, path: list[str]) -> gbt.Infoset:
```
tturocy: It makes more sense for this function to get a `node` by path - because this is a function we will eventually have as core functionality.  Then getting the infoset is trivial.   Single responsibility and decomposability!!!!

ChangeLog
```diff
@@ -2,6 +2,10 @@
 
 ## [16.5.0] - unreleased
 
+### Added
+- Implement `IsAbsentMinded()` on information sets (C++) and `Infoset.is_absent_minded` (Python)
+  to detect if an information set has absent-mindedness.
```
tturocy: Again we should be consistent with terminology.

tturocy (approved):

PR #680, opened 2025-12-08
Bump supported Python interval to [3.10, 3.14]

### LINKED ISSUE #635: MAINT: Update Python version support
Python 3.9 has reached end-of-life and Python 3.14 is out.

`master` (and by implication version 16.5) should be updated to reflect this.

It might also be worth a look to see whether there any areas where we were restricted in using syntax.  The one that comes to mind for me is that we shouldn't have to do `typing.Union` anymore (without having to do a `__future__` import).

edwardchalstrey1:
Closes #635

### DIFF
--- a/src/pygambit/action.pxi
+++ b/src/pygambit/action.pxi
@@ -108,7 +108,7 @@ class Action:
             return Rational(py_string.decode("ascii"))
 
     @property
-    def plays(self) -> typing.List[Node]:
+    def plays(self) -> list[Node]:
         """Returns a list of all terminal `Node` objects consistent with it.
         """
         return [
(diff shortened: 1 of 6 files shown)


.devcontainer/devcontainer.json
```diff
@@ -3,7 +3,7 @@
   "features": {
     "ghcr.io/devcontainers/features/python:1": {
       "installTools": true,
-      "version": "3.11"
+      "version": "3.14"
```
edwardchalstrey1, line 6: @tturocy @rahulsavani  I don't think this devcontainer is documented anywhere - do people use it? If so perhaps we should reference it in the developer/contibutor docs page

PR #682, opened 2025-12-08
Add algorithms page

### LINKED ISSUE #658: [Documentation]: Move algorithm descriptions from CLI manual pages to separate page
### Requested documentation improvement

- [ ] Create a new RST page which uses [the algorithms table](https://gambitproject.readthedocs.io/en/latest/pygambit.html#algorithms-for-computing-nash-equilibria) as a starting point
- [ ] Link to this from the homepage clearly (a 5th purple link)
- [ ] Add a description for each of the methods, adapted from the current descriptions on the CLI pages but not CLI-specific
- [ ] Update table to link to all 3 of "Description", CLI & PyGambit func, where "Description" just takes you to a header further down the page
- [ ] CLI manual info should now just be how to run the method, and there is a link to a new page for the description
- [ ] Also add links to description from PyGambit pages
 
### Consider opening as subsequent issues later

- Make each method a separately citable artefact
- Linked to #561

edwardchalstrey1:
- Closes #658 
- [New page link](https://gambitproject--682.org.readthedocs.build/en/682/algorithms.html)

### DIFF
--- a/doc/algorithms.rst
+++ b/doc/algorithms.rst
@@ -1,10 +1,10 @@
 .. _algorithms:
 
-Nash equilibria algorithms
-==========================
+Equilibrium computation
+=======================
 
-The table below summarizes the available PyGambit functions and corresponding Gambit CLI commandsto algorithms for computing Nash equilibria.
-Scroll down for detailed descriptions of each algorithm.
+The table below summarizes the available PyGambit functions and corresponding Gambit CLI programs
+for algorithms for computing Nash equilibria.
 
 ================  ===========================================================================   ========================================  ==========================================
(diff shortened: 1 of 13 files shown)


tturocy:
My first reaction is that it does expose how many gaps there are in the documentation for explaining what these methods do.  :).  (Or maybe more to the point, it exposes that there is documentation in our gaps!!!!)

In the first instance I suppose one page with subsections/anchors seems the way to go.  I would suggest our idea would be that we'd want to aim for descriptions that are no more than what would be one side of A4.  For more details we would link to original papers or surveys (where appropriate) or other sources for more information (e.g. a Jupyter notebook).  That would mean one HTML page would make sense I think.

PR #688, opened 2025-12-10
Implement postorder iterator for nodes range.

tturocy:
This extends the Nodes range iterator to have two modes, preorder (the existing one), and
postorder (added).

These are runtime-selectable via the TraversalOrder parameter. The default is preorder (so existing code still works).

### DIFF
--- a/src/games/game.h
+++ b/src/games/game.h
@@ -640,7 +640,6 @@ public:
       struct Frame {
         GameNode m_node;
         ChildIterator m_current, m_end;
-        bool m_expanded = false;
       };
 
       Game m_owner{nullptr};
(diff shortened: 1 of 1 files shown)


tturocy:
@d-kad @rahulsavani As it turned out, as I was unpicking some matters in `MixedBehaviorProfile` to work on another issue, I realised that it would be a very good idea to implement postorder iteration on nodes there as well.  So I have proposed an extension to the `Nodes` range which allows both preorder and postorder iteration.  Tests pass so the refactored preorder seems to be correct.  Postorder is not tested however.

d-kad (commented):
To test  it, I updated the game.h file in the current subgame root finder branch and replaced the suboptimal vector-reversal approach in FindSubgameRoots with a call to the new iterator. All tests in the subgames suite passed.

Issue #654, opened 2025-12-01
[Bug]: adding a "all zero payoff" outcome at a non-terminal node breaks lp_solve and lcp_solve


rahulsavani:
### Where did you find this bug?

PyGambit

### What operating system are you using?

macOS

### What version of Gambit are you using?

16.4.0

### What happened?

In the following example, which is "matching pennies" as a EFG, we get the wrong answers from `lp_solve` and `lcp_solve` by adding an outcome with payoffs `[0,0]` at a non-terminal node.

```import pygambit as gbt

def create_MP(with_neutral_outcome=False) -> gbt.Game:
    g = gbt.Game.new_tree(
        players=["Alice", "Bob"], title="Matching pennies")
    g.append_move(g.root, "Alice", ["H", "T"])
    g.append_move(g.root.children, "Bob", ["h", "t"])
    alice_lose = g.add_outcome([-1, 1], label="Alice lose")
    alice_win = g.add_outcome([1, -1], label="Alice win")
    if with_neutral_outcome:
        neutral = g.add_outcome([0, 0], label="neutral")
        g.set_outcome(g.root.children["H"], neutral)
    g.set_outcome(g.root.children["H"].children["h"], alice_win)
    g.set_outcome(g.root.children["T"].children["t"], alice_win)
    g.set_outcome(g.root.children["H"].children["t"], alice_lose)
    g.set_outcome(g.root.children["T"].children["h"], alice_lose)
    return g

g1 = create_MP()
g2 = create_MP(with_neutral_outcome=True)

for i in [0,1]:
    assert (g1.to_arrays()[i] == g2.to_arrays()[i]).all() 

print(gbt.nash.lp_solve(g1))
print(gbt.nash.lp_solve(g2))

print(gbt.nash.lcp_solve(g1))
print(gbt.nash.lcp_solve(g2))
``` 

From those solver calls we get for `lp_solve`:

```
NashComputationResult(method='lp', rational=True, use_strategic=False, equilibria=[[[[Rational(1, 2), Rational(1, 2)]], [[Rational(1, 2), Rational(1, 2)]]]], parameters={})

NashComputationResult(method='lp', rational=True, use_strategic=False, equilibria=[[[[Rational(1, 2), Rational(1, 2)]], [[Rational(1, 4), Rational(3, 4)]]]], parameters={})
```

and for `lcp_solve`:

```
[...]

PR #698, opened 2025-12-14
Lp lcp

StephenPasteris:
Updated LP and LCP solvers, in doing so fixing a bug involving internal and missing outcomes (Issue #654). Sequence form object has been updated so that payoffs are on terminal nodes. Sequence form is now an attribute of a Game object. In addition, several tests have been added.

### DIFF
--- a/tests/test_nash.py
+++ b/tests/test_nash.py
@@ -143,6 +143,21 @@ def test_enummixed_rational(game: gbt.Game, mixed_strategy_prof_data: list):
                  [[0.3333333333333333, 0.6666666666666667], [1.0, 0.0]]]],
             2,
         ),
+        (
+            games.create_entry_accomodation_efg(),
+            [
+                [[[0.6666666666666666, 0.33333333333333337], [1.0, 0.0], [1.0, 0.0]],
+                 [[0.6666666666666666, 0.33333333333333337]]],
+                [[[0.0, 1.0], [0.0, 0.0], [0.3333333333333333, 0.6666666666666667]], [[0.0, 1.0]]],
+                [[[0.0, 1.0], [0.0, 0.0], [1.0, 0.0]], [[1.0, 0.0]]],
+                [[[0.0, 1.0], [0.0, 0.0], [0.0, 0.0]], [[0.0, 1.0]]]],
+            4,
+        ),
+        (
+            games.create_non_zero_sum_lacking_outcome_efg(),
+            [[[[0.33333333333333337, 0.6666666666666666]], [[0.5, 0.5]]]],
+            1,
+        ),
     ],
 )
 def test_enumpoly_ordered_behavior(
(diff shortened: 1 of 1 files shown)


src/pygambit/gambit.pyx
```diff
@@ -50,7 +50,10 @@ class Rational(fractions.Fraction):
 @cython.cfunc
 def rat_to_py(r: c_Rational):
     """Convert a C++ Rational number to a Python Rational."""
-    return Rational(to_string(r).decode("ascii"))
```
tturocy, line 53: I'm not sure I like this change.  A rational will never decode to an empty string, so if this happens there is an error somewhere.  This change will just hide that error.

src/games/game.h
```diff
@@ -965,6 +1003,49 @@ class GameRep : public std::enable_shared_from_this<GameRep> {
 
   /// Build any computed values anew
   virtual void BuildComputedValues() const {}
+
+  // Sequence form methods
+
+  virtual std::shared_ptr<BehaviorSupportProfile> GetFullSupport()
+  {
+    throw std::runtime_error("Sequence form can only be generated for extensive form games");
```
tturocy, line 1011: We already have a not implemented exception which is used for this purpose.

src/games/gametree.h
```diff
@@ -47,7 +47,48 @@ class GameTreeRep : public GameExplicitRep {
   std::map<GameNodeRep *, std::vector<GameNodeRep *>> m_nodePlays;
   mutable std::shared_ptr<OwnPriorActionInfo> m_ownPriorActionInfo;
   mutable std::unique_ptr<std::set<GameNodeRep *>> m_unreachableNodes;
-  mutable std::set<GameInfosetRep *> m_absentMindedInfosets;
+  mutable std::shared_ptr<BehaviorSupportProfile> m_fullSupport;
+
+  std::shared_ptr<BehaviorSupportProfile> GetFullSupport() override
```
tturocy, line 52: This is not the way to do this.  We should be cacheing a sequence form, not a full support.

src/solvers/lcp/efglcp.cc
```diff
@@ -237,107 +237,87 @@ void NashLcpBehaviorSolver<T>::AllLemke(const Game &p_game, int j, linalg::Lemke
...
-  if (n->GetPlayer()->IsChance()) {
-    for (const auto &action : infoset->GetActions()) {
-      FillTableau(A, n->GetChild(action),
-                  Rational(prob) * static_cast<Rational>(infoset->GetActionProb(action)), s1, s2,
-                  p_solution);
+  auto player1 = p_game->GetPlayer(1);
+  auto player2 = p_game->GetPlayer(2);
+  auto sequences1 = p_game->GetSequences(player1);
+  auto sequences2 = p_game->GetSequences(player2);
+  for (auto seq : sequences1) {
```
tturocy, line 250: This is definitely where the refactoring work is starting to pay off, as this is getting closer to being much easier to read/understand!

src/games/gameseq.cc
```diff
@@ -67,22 +67,26 @@ void GameSequenceForm::BuildSequences()
 }
 
 void GameSequenceForm::FillTableau(const GameNode &n, const Rational &prob,
```
tturocy, line 69: We definitely need to deal with this recursion and convert it to a stack-based method.  But that is outside the scope of this task.

src/games/game.cc
```diff
@@ -409,4 +410,74 @@ template class MixedStrategyProfileRep<Rational>;
...
+                         m_support->GetSequenceForm()->m_sequences.cend(), 0,
+                         [](int acc, const std::pair<GamePlayer, std::vector<GameSequence>> &seq) {
+                           return acc + seq.second.size();
+                         });
+}
+
+Sequences::iterator Sequences::begin() const { return {m_support->GetSequenceForm(), false}; }
+Sequences::iterator Sequences::end() const { return {m_support->GetSequenceForm(), true}; }
+
+Sequences::iterator::iterator(const std::shared_ptr<GameSequenceForm> p_sfg, bool p_end)
```
tturocy, line 429: We have a `NestedElementCollection` which implements exactly this generically - see how `Infosets` and `Strategies` are implemented.

Issue #696, opened 2025-12-12
[Bug]: UndefinedOperationError not thrown for game.Player.actions on non-tree game


rahulsavani:
### Where did you find this bug?

PyGambit

### What operating system are you using?

macOS

### What version of Gambit are you using?

16.4.0

### What happened?

If one has a non-tree game `g` and does `g.actions` we throw a `UndefinedOperationError`.

It seems logical to throw the same error if one calls actions for a player in this game, but instead we just return a empty set of `PlayerActions`, as in the following:

```
In [1]: import pygambit as gbt

In [2]: g = gbt.read_nfg('tests/test_games/2x2x2_nfg_with_two_pure_one_mixed_eq.nfg')

In [3]: g.actions
---------------------------------------------------------------------------
UndefinedOperationError                   Traceback (most recent call last)
Cell In[3], line 1
----> 1 g.actions

File ~/gambit/src/pygambit/game.pxi:701, in pygambit.gambit.Game.actions.__get__()
    699 """
    700 if not self.is_tree:
--> 701     raise UndefinedOperationError(
    702         "Operation only defined for games with a tree representation"
    703     )

UndefinedOperationError: Operation only defined for games with a tree representation

In [4]: g.players[0].actions
Out[4]: PlayerActions(player=Player(game=Game(title='2x2x2 game with 2 pure and 1 mixed equilibrium'), label='Player 1'))

In [5]: len(g.players[0].actions)
Out[5]: 0
```

PR #701, opened 2025-12-15
UndefinedOperationError for game.player.{actions,infosets} on non-tree

rahulsavani:
closes issue #696


tturocy (changes requested):
Per comment - add this check to `nodes` as well please.  Other functions looks like they're fine.

PR #705, opened 2025-12-16
Added tests for enumpoly, LP and LCP

StephenPasteris:
Added tests for enumpoly, LP and LCP

### DIFF
--- a/tests/test_nash.py
+++ b/tests/test_nash.py
@@ -418,20 +418,23 @@ def test_lcp_behavior_double():
                 games.create_2x2_zero_sum_efg(),
                 [[["1/2", "1/2"]], [["1/2", "1/2"]]]
         ),
-        (
+        pytest.param(
             games.create_2x2_zero_sum_efg(missing_term_outcome=True),
             [[["1/2", "1/2"]], [["1/2", "1/2"]]],
+            marks=pytest.mark.xfail(reason="Problem with non-standard outcomes")
         ),
         (games.create_matching_pennies_efg(),
          [[["1/2", "1/2"]], [["1/2", "1/2"]]]),
-        (
+        pytest.param(
             games.create_matching_pennies_efg(with_neutral_outcome=True),
             [[["1/2", "1/2"]], [["1/2", "1/2"]]],
+            marks=pytest.mark.xfail(reason="Problem with non-standard outcomes")
         ),
         (games.create_stripped_down_poker_efg(), [[[1, 0], ["1/3", "2/3"]], [["2/3", "1/3"]]]),
-        (
+        pytest.param(
             games.create_stripped_down_poker_efg(nonterm_outcomes=True),
             [[[1, 0], ["1/3", "2/3"]], [["2/3", "1/3"]]],
+            marks=pytest.mark.xfail(reason="Problem with non-standard outcomes")
         ),
         (
                 games.create_kuhn_poker_efg(),
@@ -447,7 +450,7 @@ def test_lcp_behavior_double():
                     [[1, 0], ["2/3", "1/3"], [0, 1], [0, 1], ["2/3", "1/3"], [1, 0]],
                 ],
         ),
-        (
+        pytest.param(
             games.create_kuhn_poker_efg(nonterm_outcomes=True),
             [
                 [
@@ -460,13 +463,10 @@ def test_lcp_behavior_double():
                 ],
                 [[1, 0], ["2/3", "1/3"], [0, 1], [0, 1], ["2/3", "1/3"], [1, 0]],
             ],
+            marks=pytest.mark.xfail(reason="Problem with non-standard outcomes")
         ),
         # In the next test case:
         # 1/2-1/2 for l/r is determined by MixedBehaviorProfile.UndefinedToCentroid()
-        (
-                games.create_two_player_perfect_info_win_lose_efg(),
-                [[[0, 1], [1, 0]], [[0, 1], ["1/2", "1/2"]]],
-        ),
         (
                 games.create_perfect_info_with_chance_efg(),
                 [[[0, 1]], [[0, 1], [0, 1]]],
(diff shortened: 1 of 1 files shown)


tests/test_nash.py
```diff
@@ -393,30 +445,78 @@ def test_lcp_behavior_double():
...
+        ),
+        pytest.param(
+                games.create_one_card_poker_lacking_outcome_efg(),
+                [[["2/3", "1/3"]], [[1, 0], ["1/3", "2/3"]]],
+                marks=pytest.mark.xfail(reason="Problem with missing terminal outcome")
+        ),
+        pytest.param(
+                games.create_perfect_info_internal_outcomes_efg(),
+                [[[0, 1], [1, 0]], [[0, 1], ["1/2", "1/2"]]],
+                marks=pytest.mark.xfail(reason="Problem with internal outcomes")
```
rahulsavani: This case that starts on line 465 actually xpasses rather than xfails. If you are happy with this test as being useful then by all means keep it but remove the the xfail mark. If you think a similar test to this one can be made to xfail, add that variant.

tests/games.py
```diff
@@ -123,6 +123,230 @@ def create_2x2_zero_sum_efg(missing_term_outcome: bool = False) -> gbt.Game:
...
+        g.root.children[0].children[1].children[0], g.add_outcome([-2, 2], label="aRC")
+    )
+    g.set_outcome(
+        g.root.children[0].children[1].children[1], g.add_outcome([-2, 2], label="aRD")
+    )
+    g.set_outcome(g.root.children[1], g.add_outcome([-1, 1], label="b"))
+    return g
+
+
+def create_one_card_poker_lacking_outcome_efg() -> gbt.Game:
```
rahulsavani: You have based this on old code. See PR #608.

We only have "stripped_down_poker" in the test suite and don't talk about Myerson at all, but Reiley et al.

Moreover we already create two versions of this game in the test suite, see:

`create_stripped_down_poker_efg(nonterm_outcomes: bool = False)`

rahulsavani (changes requested):
The new games have been added without being consistent with what is already there:

1. a version of stripped down poker is created in a new function in `tests/games.py` but that new code still refers to Myerson not Reiley; moreover it ignores the existing `create_stripped_down_poker_efg`

2. the new games added with non-terminal or missing outcomes etc. do not have variants with all terminal outcomes; note that the existing functions from Master for the relevant tests with and without non-terminal outcomes (`create_stripped_down_poker_efg`, `create_matching_pennies_efg`, and `create_kuhn_poker_efg`) all have an argument that allows to create a standard version with only (and all) terminal outcomes and a non-standard one. The tests then include a case for the standard version, and a xfail variant created via the same function in game with an appropriate argument. 

Here's an example:

```
        (
                games.create_2x2_zero_sum_efg(),
                [[["1/2", "1/2"]], [["1/2", "1/2"]]]
        ),
        pytest.param(
            games.create_2x2_zero_sum_efg(missing_term_outcome=True),
            [[["1/2", "1/2"]], [["1/2", "1/2"]]],
            marks=pytest.mark.xfail(reason="Problem with missing terminal outcomes")
        ),
```

Here's another example:

```
        (games.create_matching_pennies_efg(),
         [[["1/2", "1/2"]], [["1/2", "1/2"]]]),
        pytest.param(
            games.create_matching_pennies_efg(with_neutral_outcome=True),
            [[["1/2", "1/2"]], [["1/2", "1/2"]]],
            marks=pytest.mark.xfail(reason="Problem with missing outcomes")
        ),
```

Here's a final two examples where we are not xfailing because the existing code already gives the right answer even with non-terminal outcomes:

```
        (games.create_stripped_down_poker_efg(), [[["1/3", "2/3", 0, 0], ["2/3", "1/3"]]], None),
        (
                games.create_stripped_down_poker_efg(nonterm_outcomes=True),
                [[["1/3", "2/3", 0, 0], ["2/3", "1/3"]]],
                None
        ),
        (games.create_kuhn_poker_efg(), [games.kuhn_poker_lcp_first_mixed_strategy_prof()], 1),
        (
                games.create_kuhn_poker_efg(nonterm_outcomes=True),
                [games.kuhn_poker_lcp_first_mixed_strategy_prof()],
                1
        ),
```

In these examples the two versions, with and without non-terminal outcomes, gave the same answer (hence e.g. `poker_lcp_first_mixed_strategy_prof()` is used in both tests. I believe you have an example where making such a variation changed the output of the LP or LCP solver. Clearly we want both tests in that case, which is interesting. But actually hopefully it is clear that it always makes sense to have a test for both variants.

I presume you put the "expected outputs" for the xfailing variants based on a standard variant anyhow? Or did you already have working code for these xfailing variants? Either way, please add tests for the standard variants as well.

For reconciling `create_one_card_poker_lacking_outcome_efg()` with `create_stripped_down_poker_efg` it seems that you might want to add a third variant. In that case you might want to go from a boolean flag in to some kind of "type" argument, which if `None` would give the standard one.

tests/test_nash.py
```diff
@@ -135,10 +135,62 @@ def test_enummixed_rational(game: gbt.Game, mixed_strategy_prof_data: list):
...
+        pytest.param(
+            games.create_large_payoff_game_efg(),
+            [[[[1.0, 0.0], [1.0, 0.0]], [[0.0, 1.0], [1.0, 0.0]]],
+             [[[1.0, 0.0], [1.0, 0.0]], [[0.0, 1.0], [0.0, 0.0]]],
+             [[[1.0, 0.0], [0.0, 1.0]], [[0.0, 1.0], [1.0, 0.0]]],
+             [[[0.0, 1.0], [1.0, 0.0]], [[0.0, 1.0], [1.0, 0.0]]],
+             [[[0.0, 1.0], [1.0, 0.0]], [[0.0, 1.0], [1.0, 0.0]]],
+             [[[0.0, 1.0], [0.0, 1.0]], [[0.0, 0.0], [1.0, 0.0]]]],
+            6,
+            marks=pytest.mark.xfail(reason="True equilibrium has tiny probabilities")
```
rahulsavani: I don't follow: You say the true equilibrium has tiny probabilities but all the expected equilibria above are 0.0 or 1.0 (are they the true equilibria or not)? 

Does this fail if the `max_regret` check is removed?

tests/games.py
```diff
@@ -123,6 +123,233 @@ def create_2x2_zero_sum_efg(missing_term_outcome: bool = False) -> gbt.Game:
...
+
+def create_large_payoff_game_efg() -> gbt.Game:
+    g = gbt.Game.new_tree(players=["1", "2"], title="Large payoff game")
+    g.append_move(g.root, g.players.chance, ["L", "R"])
+    for i in range(2):
+        g.append_move(g.root.children[i], "1", ["A", "B"])
+    for i in range(2):
+        g.append_move(g.root.children[0].children[i], "2", ["X", "Y"])
+        g.append_infoset(g.root.children[1].children[i], g.root.children[0].children[i].infoset)
+    o_large = g.add_outcome([10000000000000000000, -10000000000000000000], label="large payoff")
```
rahulsavani, line 295: This is a huge payoff. Make the numbers more reasonable and we should be able to get `enumpoly` giving us suitable max regret and matching our expected exact equilibria.

PR #711, opened 2025-12-16
Add `GetPlayersWithChance()` range

tturocy:
This adds to `GameRep` a new `GetPlayersWithChance()`, which allows transparent iteration over the chance player as the "first" player in the collection of players.


d-kad:
the changes look good -- will refactor the subgames algorithm using these when pushed

PR #722, opened 2025-12-19
Lemke on trees stops after finding 1 eq; stop_after/max_depth not used

### LINKED ISSUE #671: [Bug]: Lemke (via lcp_solve) in mixed behaviors seems to only ever find on equilibrium

Other

macOS

16.4.0

### What happened?

For an EFG, `lcp_solve` uses Lemke-Howson (LH) on the strategic form and Lemke on the sequence form. For both, the docs currently claim that they will find all "accessible" equilibria.

For LH, this seems to work fine. E.g. on an nxn coordination game we find all (2^n)-1 equilibria. For Lemke, it is not really clear how to define a scheme that would find all accessible equilibria (for LH, the choice is which pure strategy to drop; for Lemke, one makes the non-discrete choice of a covering vector). In any case, I have seen no example (from many) where our implementation of Lemke actually finds more than one equilibrium.

Surely Lemke can be used to find more than one, and we should return to this in due course -- see issue #370.

For now, we should remove `stop_after` as a parameter for Lemke (`lcp_solve` on an EFG with `use_strategic` as False), and stop after we find one equilibrium, updating the docs accordingly.

rahulsavani:
closes #671

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -9,7 +9,7 @@
 ### Added
 - Implement `IsAbsentMinded()` on information sets (C++) and `Infoset.is_absent_minded` (Python)
   to detect if an information is absent-minded.
-- Tests for EFG Nash solvers -- `enumpoly_solve`, `lp_solve`, `lcp_solve` -- in behavior stratgegies
+- Tests for EFG Nash solvers -- `enumpoly_solve`, `lp_solve`, `lcp_solve` -- in behavior strategies
 - In `pygambit`, `Node` objects now have a read-only property `own_prior_action` and `Infoset` objects
   have a read-only property `own_prior_actions` to retrieve the last action or the set of last actions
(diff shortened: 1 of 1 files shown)


src/solvers/lcp/efglcp.cc
```diff
@@ -144,98 +142,23 @@ std::list<MixedBehaviorProfile<T>> NashLcpBehaviorSolver<T>::Solve(const Game &p
...
+    solution.m_equilibria.push_back(profile);
+    this->m_onEquilibrium(profile, "NE");
   }
   catch (std::runtime_error &e) {
     std::cerr << "Error: " << e.what() << std::endl;
   }
   return solution.m_equilibria;
 }
 
-//
```
tturocy: We're confident enough that this is pure nonsense that we should be rid of it entirely rather than guarding it with `#ifdef UNUSED` or similar for now?

PR #723, opened 2025-12-23
Automatically normalise information set ordering on demand

tturocy:
This implements normalisation of the iteration order of information sets per player on demand.
Previously `sort_infosets` needed to be called to ensure this.  However, information set sorting could also be
triggered implicitly, for example, by computing the strategies for the game, leading to unexpected side effects.

`sort_infosets` is now deprecated and will be removed in a future version.  It remains as a no-op for compatibility.


rahulsavani (approved):
Looks good to me.

PR #736, opened 2026-01-14
Clean up tests/games.py

rahulsavani:
- Removed many functions from games.py that were either simply reading a file or were creating a game but the creation code was otherwise uninteresting for us; instead for these games we use `games.read_from_file` to read .efg or .nfg files directly in the tests themselves.
- Used `games.create_efg_corresponding_to_bimatrix_game` to simplify two functions.

Some of the new EFGs could do with better titles and descriptions but that applies to many old ones too, so I left that for further work.


rahulsavani:
This work has been incorporated in PR #724 .

PR #750, opened 2026-02-01
Implement efficient subgame root detection via interval reachability

### LINKED ISSUE #584: FEAT: Implement improved algorithm for finding subgame roots

d-kad:
**Description**

This PR implements efficient subgame root detection for extensive-form games using an adaptation of Tarjan's bridge-finding algorithm.

**Algorithm overview:**

A node is a subgame root iff no information set straddles the subtree rooted at this node. 
The implementation uses a two-phase linear approach:

1. **Subtree Spans and Infoset Hulls**: Number terminal nodes in post-order; propagate intervals up the tree and compute the hull of intervals for each information set
2. **Bridge Detection**: Compare reachable span and subtree span of each node (cf. Tarjan's interval bridge-finding condition)

**C++ changes (gametree.cc, gametree.h):**
- Add `BuildSubgameRoots()` implementing the interval reachability algorithm
- Add `GetSubgames()` returning subgame roots in post-order (deepest first)
- Refactor `IsSubgameRoot()` to use cached results with lazy computation

**Test changes:**
- Refactor `test_subgame_roots` to use factory pattern and dataclasses
- Add test games for non-trivial subgames with nested and overlapping infosets (with and without Nature)

**Notes:**
- `IsSubgameRoot()` lookup is currently O(S), where S is the number of subgames; will be refactored to O(1) lookups when `GameSubgameRep` is implemented

Closes #584

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -6,8 +6,9 @@
 - `Game.comment` has been renamed to `Game.description`
 
 ### Added
-- Implement `GetSubgames()` (C++) to retrieve the list of subgame roots in standard post-order.
-  Detection uses a linear interval reachability algorithm adapted from the Tarjan's bridge-finder (IPL, 1974).
+- Implement linear-time algorithm to find all root nodes of proper subgames, using an adaptation of
+  Tarjan's (1974) algorithm for finding bridges in an undirected graph.  Subgame roots are cached so
+  subsequent lookup is constant-time (if the game is unchanged).  (#584)
 
 ### Fixed
 - `enumpoly` would take a very long time on some supports where an equilibrium is located on the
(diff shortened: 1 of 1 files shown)


src/games/gametree.cc
```diff
@@ -1139,6 +1128,126 @@ void GameTreeRep::BuildUnreachableNodes() const
...
+    return;
+  }
+
+  struct Range {
+    int min = std::numeric_limits<int>::max();
+    int max = 0;
+
+    void Merge(const Range &p_source)
+    {
+      if (p_source.min < min) {
```
tturocy: `std::min` and `std::max`!!!

PR #753, opened 2026-02-03
Add initial version of games catalog

### LINKED ISSUE #731: Create initial games catalog from contrib/games

PyGambit

This issue is meant to scope an initial PR of the games library/catalog idea discussed in #623 but not yet to move all tests & documentation/tutorials etc over to loading from the catalog.

To do:
- [x] Create a new folder called `/catalog` and move the EFG/NFG files from contrib/games there
- [x] Create a new `catalog.py` which contains the catalog of games.
- [x] Decide whether to structure catalog and contained games as a class with functions, class with subclasses, or [dataclasses](https://docs.python.org/3/library/dataclasses.html). You should be able to get a game like `g = pygambit.catalog.PrisonersDilemma()`
- [x] Should be a way to retrieve/list all games
- Each game needs a name that makes sense according to a consistent naming convention that can easily remain stable
    - [x] Derive class name slugs from the filename for now
    - [x] Restore getting `game_type` from `game_file`
    - [x] Restore loading attributes from games - no metadata fields should be added that can already be part of a game object anyway
        - But we still want ability to add metadata e.g. is_perfect_recall 
    - [x] Get the description field from the gambit game "comment"
        - it should let you do arbitrary text
        - Citation should be included in comment
    - [x] Decide a way to auto-generated from contrib files
    - [x] Write a script that builds a "database" from contrib/games by updating `catalog.py`
        - [x] Make it so this script ignores files that already have a class in the catalog, just adding anything not already added
        - [x] Ensure comments are retained
        - Ted and Rahul can then go in and update the "database" for the slug/class, title & comment/description
 - Think about whether `catalog.py` should be refactored to be a proper database, whether a yml/json file or sqlite etc
    - ~~I think a json/yml file would be harder to maintain and look at diffs in PRs, db is overkill~~
    - [x] Add `catalog.yml` which is generated by the database building script
    - [x] Make it so we can add whatever custom metadata we want to each class -  It could be that we update the Game class to have an attribute which is a dataclass for the catalog metadata
        - ~~Another option could be a "catalog metadata" dataclass that is added to Game objects directly~~
        - [x] Add catalog metadata to classes
        - [x] Ignore any that have `valid_game: false`
        - [x] You should be able to filter on all metadata fields, whether they come from the game object or the catalog metadata
- [x] Each game has to have the following attributes:
    - [x] title
    - [x] description
    - [x] citation
    - [x] no. of players
        - [x] should be able to filter on this
    - [x] extensive or normal form
        - [x] should be able to filter on this
- [x] For the initial games (from contrib/games) the functions will simply call `read_efg("game.efg")` or `read_nfg("game.nfg")` and return the game object 
    - [x] Add a class for all the efg and nfg games
- [x] Add tests
- [x] Ensure catalog functions are properly documented in API reference
    - [x] Make it so that all classes that are generated from `catalog.yml` have a docstring, which gets populated by `game.comment` or  description field (not metadata) in the catalog entry
- [x] Some of the game functions (or classes) you can pass in parameters to get variants of games
- [x] New games can be added to the catalog as functions that create the game from the python code directly rather than loading from efg/nfg e.g. those currently in `tests/games.py`
- [x] Ensure that `update.py` adds games to the rst doc from catalog.py as well as catalog.yml
- [x] Refactor so that metadata extraction from game object to catalog entry is no longer needed and the `games()` function builds each game when it runs
- [x] Refactor so that catalog description overwrites game description rather than the other way around
- [x] Update games() to use all possible attributes of Game objects when filtering and to use `subclass.game` to get them
- [x] Make it so the notebook tests don't fail with an `ImportError` when doing `from pygambit import catalog`
    - Should we implement a `catalog.load()` function which generates the classes, so we don't have to do so when importing pygambit to begin with?
- [x] Consider whether to move manually coded files back to their own script or keep in `catalog.py`
- [x] Ensure `catalog_update.py` works as expected now catalog location moved
- [x] `catalog_update.py` should also update `Makefile.am` OR the latter should load from `catalog.yml`
- [x] If the filename already exists in the catalog but the class has been renamed, then running `catalog_update.py` again should update rst
- [x] Make it so that you can provide any custom metadata in catalog.yml with a `docstring` field that gets used to describe the field in the docs page - can we make each metadata field an optional attribute of `CatalogGame` with a type and possible values
- [x] Rename `contrib/games` to `catalog_game_files`
- [x] Update docs with links to contrib files
- [x] Refactor how the _game func works in coded games such that args are displayed in the docs
    - [x] Why is `W_ex1` and similar catalog funcs not getting docs built with _game docstring?
loaded from (could have an `img` field in `catalog.yml`)
- [x] Move the catalog API reference to an easy place to find or link to it and update `catalog_update.py`
- [x] Test the `coded_games` and make sure an error is thrown if you try to make one with the `_game` function not returning a game
- [x] Add a test that validates that all entries in catalog.yml are valid
- Add documentation for how to update the catalog including
    - [x] adding to `catalog.yml`
    - [x] adding to `catalog_games.py`
    - [x] how the `catalog_update.py` script works
    - [ ] that you also need to manually update links to that game in any documentation **if** you rename the file 
    - [x] `pip install ruamel.yaml`
    - [x] GUI -> EFG -> catalog pipeline
    - [x] How to set up a metadata filter, giving a default on `CatalogGame`
- [x] Update tutorials with clear instructions on how to use the catalog, maybe just in one place rather than two and ensure all examples are using the catalog rather than loading from file
- [x] Ensure game description does not have full docstring
- [x] Can we build the docstring including parts derived from a game object when generating docs with make html?
    - [x] Can we also include game title as part of the catalog game class docstring?

edwardchalstrey1:
### Issues closed by this PR

- Closes #731

This PR adds infrastructure and documentation for a first version of the Catalog of games:
- Catalog games can be created by adding NFG/EFG files to `catalog` dir
- pygambit can load games from outside the pygambit code (in `catalog` dir)
- Moves several game files from `contrib/games` to `catalog` to test catalog functionality
- Updates tutorials to load from catalog
- Adds a page to the docs with all Catalog games, linked to on the docs navbar which just has the game slug and title for now
- Adds a script called `src/pygambit/update_catalog.py` that you run after adding a new EFG/NFG (usage described in developer docs)
- Also contains change from #757 

### How to review this PR

- Look at the [Catalog of games page](https://gambitproject--753.org.readthedocs.build/en/753/catalog.html)
- Review the developer docs page on [Updating the Games Catalog](https://gambitproject--753.org.readthedocs.build/en/753/developer.catalog.html)
- Take a look at the [Loading games from the catalog](https://gambitproject--753.org.readthedocs.build/en/753/tutorials/01_quickstart.html#Loading-games-from-the-catalog) section added to Tutorial 1

### DIFF
--- a/tests/test_catalog.py
+++ b/tests/test_catalog.py
@@ -31,4 +31,4 @@ def test_catalog_games():
     assert isinstance(all_games, pd.DataFrame)
     assert len(all_games) > 0
     assert "myerson1991/fig4_2" in list(all_games.Game)
-    assert "Myerson (1991) Fig 4.2" in list(all_games.Title)
+    assert "Myerson (1991) Figure 4.2" in list(all_games.Title)


pyproject.toml
```diff
@@ -87,5 +88,12 @@ markers = [
     "slow: all time-consuming tests",
 ]
 
+[tool.setuptools]
+packages = ["pygambit", "pygambit.catalog"]
+package-dir = { "pygambit" = "src/pygambit", "pygambit.catalog" = "catalog" }
```
edwardchalstrey1, line 108: This is how we get the `pygambit.catalog` module to be built using code from `catalog` dir at the root of the repo

pyproject.toml
```diff
@@ -87,5 +88,12 @@ markers = [
...
+
+[tool.setuptools.package-data]
+"pygambit.catalog" = ["*"]
```
edwardchalstrey1, line 111: This is how the catalog files get included as package data

edwardchalstrey1:
@tturocy @rahulsavani this PR can be merged once the game slugs (filenames) and titles for the 4 examples have been updated, i.e. inside the EFG/NFG files in `/catalog`.

Note that I used the game titles in a few places in `test_catalog.py`, so will might need to update there too if any are changed.

PR #757, opened 2026-02-05
Fix printing normal form games

edwardchalstrey1:

@tturocy I noticed that after recent changes when you view a normal form game in jupyter the printed html is duplicated:

Screenshot 2026-02-05 at 09 45 57

I've made a temporary fix with out looking properly into how the code works and now we get the game printed just once:
Screenshot 2026-02-05 at 09 48 07

### How to review this PR

@tturocy I can take more of a look into this but you may know a quick fix


edwardchalstrey1:
@tturocy Thanks for the `git bisect` suggestion. Not used that before, very useful!

Looks like the problem commit was: https://github.com/gambitproject/gambit/commit/fce47661a1f2c8a4269d3a425823b4d54b485418

PR #766, opened 2026-02-13
Update graphical interface with new branding

tturocy:
This updates the graphical interface to incorporate our new logo:

* Updates the MacOS .icns file for the app
* Refines the splash-screen experience to be a bit more modern
* Updates and simplifies the "About" dialog

These have been tested so far on MacOS - we need to review these on Linux and MSW.  Specifically we need to see whether there's any adjustments to the MSW installer that can/should be made to incorporate the branding.

### DIFF
--- a/src/gui/dlabout.cc
+++ b/src/gui/dlabout.cc
@@ -76,7 +76,6 @@ AboutDialog::AboutDialog(wxWindow *p_parent)
 
   auto *panel = new wxPanel(this);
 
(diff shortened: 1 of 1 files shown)


tturocy:
@rahulsavani please try on Linux; @edwardchalstrey1 if we can manage to scare up a Windows machine somewhere to try this on it would be useful (there should be an installer built as an artefact on the GitHub action on the branch).

PR #774, opened 2026-02-20
Refactor apis of mixed profiles to be products of distributions (#763)

### LINKED ISSUE #763: [ENH] Revise API of mixed strategy/behaviour profiles to be products of distributions
At present, the interface for `MixedStrategyProfile`/`MixedBehaviorProfile` (and their Python equivalents) allow the setting of a probability of an individual strategy/action.  Based on some discussions, it seems that this is probably not an ideal way to express these objects.

Both of these are actually products of multinomial probability distributions.  It is difficult to think of a legitimate use case in which the probability of just one action/strategy would need to be set, as opposed to setting the whole vector of probabilities.

We do support counts/weights data for the components of these profiles.  This is useful as a convenience, but also it is essential for estimation - for example the counts are needed as inputs to `logit_estimate` in order to return the correct log-likelihoods.

This therefore suggests the following:
* We have a concept of a "multinomial distribution".  In our case, we would allow weights/counts, similar to Python's `random.choices`. for example.  So any distribution of non-negative numbers would be valid, *except* for all zeroes.
* We remove setters that set the probability of an individual strategy or action, and set only the distribution over a player or an information set.  As our current (and intended) API treats strategies and actions as being ordered, this is not ambiguous (although we might also allow a mapping-like objects of labels to values as well in Python).
* We retain the existing `normalize` operation for explicit normalisation.
* For computing expected payoffs, weights would be implicitly normalised to proper probability distributions.  

Perhaps the open question is whether to have different setters for a proper probability distribution versus weights, or to have a separate concept of mixed profiles versus "weight" profiles - where the former would insist on sum-to-one as a way to help calling code avoid subtle bugs if it intends to have probability distributions but for some reason does not.

hardik-xi11:
- Closes #763

- refactored python apis for MixedStrategyProfile and MixedBehaviorProfile to correctly reflect their structure as products of distributions.

- before, the api allowed setting the probability of individual strategies or actions. i have removed those individual setters and enforced distribution-level setting per player or per information set.

### DIFF
--- a/src/pygambit/behavmixed.pxi
+++ b/src/pygambit/behavmixed.pxi
@@ -130,40 +130,11 @@ class MixedAction:
                 raise KeyError(f"no action with label '{index}' at infoset") from None
         raise TypeError(f"strategy index must be Action or str, not {index.__class__.__name__}")
 
-    def __setitem__(self, index: ActionReference, value: typing.Any) -> None:
-        """Sets the probability an action is played.
-
-        Parameters
-        ----------
-        index : Action or str
-            The part of the profile to set:
-
-            * If `index` is an ``Action``, sets the probability the action is played.
-            * If `index` is a ``str``, attempts to resolve the referenced object by searching
-              for an action with that label, and sets the probability for that action.
-
-        value
-            Any value which can be converted to the data type of the ``MixedBehaviorProfile``.
-
-        Raises
-        ------
-        MismatchError
-            If `action` is an ``Action`` that does not belong to this ``MixedAction``'s
-            information set.
-        """
-        self.profile._check_validity()
-        if isinstance(index, Action):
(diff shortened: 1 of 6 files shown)


tturocy:
We are not able to review unsolicited pull requests that have not been previously discussed with a developer.

If you are interested in participating in GSoC 2026, please read https://www.gambit-project.org/gsoc_2026/ and specifically "Contributor Guidance".  With that information we can assess whether you are likely a good candidate to work on Gambit and provide with suggestions and guidance on initial issues to try out.

As a specific note, the issue states this is a C++ matter, and it is.  There is a need to create a concept of a probability distribution and use throughout the codebase.  This issue would take me the better part of a week to complete adequately - for someone new to the codebase it would take a summer.

PR #779, opened 2026-02-21
Fix: Resolve undefined off-path probabilities in enumpoly and drop fake equilibria (#660)

### LINKED ISSUE #660: [Bug]: enumpoly returns a profile with undefined probabilities that may not be a Nash eq

PyGambit

macOS

16.4.0

### What happened?

This was spotted when working on issue #636, which is about using `UndefinedToCentroid` rather than leaving probabilities at certain infosets as 0 in pygambit, or *s in the GUI. In making that change I noticed the following case (currently incorrectly included as a test case in `test_enumpoly_ordered_behavior` in `test_nash.py`).

The game in question is the "one-shot trust game", created in `tests/games.py::create_one_shot_trust_efg`, with the following efg file:

```
EFG 2 R "One-shot trust game, after Kreps (1990)" { "Buyer" "Seller" }
""

p "" 1 1 "" { "Trust" "Not trust" } 0
p "" 2 1 "" { "Honor" "Abuse" } 0
t "" 1 "Trustworthy" { 1, 1 }
t "" 2 "Untrustworthy" { -1, 2 }
t "" 3 "Opt-out" { 0, 0 }
```

The issue is that the second "equilibrium" found by `enum_poly` in mixed behaviors is not always an equilibrium:

```
import pygambit as gbt
g = gbt.read_efg('one_shot_trust.efg')
result = gbt.nash.enumpoly_solve(g)
print(result.equilibria[0])
print(result.equilibria[1])
```

giving

```
In [14]: print(result.equilibria[0])
[[[0.0, 1.0]], [[0.5, 0.5]]]

In [15]: print(result.equilibria[1])
[[[0.0, 1.0]], [[0.0, 0.0]]]
```

The second once should have `[0.0, 1.0]` for player 2, but is instead returned as "undefined", i.e. `[0.0, 0.0]` in pygambit, and wildcards * in the GUI:

Image

However, if player 2 puts more than 0.5 probability on "Honor" then "No Trust" is no longer a best response for player 1. 

Player 2's info set is only one deviation from the equilibrium path (i.e. "Not Trust"), and so actually should have specified probabilities and not be undefined (which should occur only for info sets that would only be reached with deviations of more than one player).

satyam102006:
### Related Issue
Resolves #660 

### Description of the Problem
The `enumpoly_solve` function was returning invalid and duplicate equilibria with undefined `[0, 0]` probabilities at off-path information sets (most notably in the One-Shot Trust Game and its unique variant). 

Attempting to assign probabilities directly to the sparse solution array during Nash extension checks was also causing `Index out of range` errors when passing back to Cython.

### What this PR does
**1. C++ Core (`src/solvers/enumpoly/efgpoly.cc`)**
* **Safe Memory Allocation:** Initializes a dense `MixedBehaviorProfile` (`full_sol`) for the entire game tree and copies valid probabilities over. This prevents Cython `Index out of range` crashes when testing off-path actions.
* **Strict Nash Verification:** The solver now actively identifies undefined off-path infosets (where the sum of probabilities is `< 0.5`). It systematically tests assigning pure strategies (`1.0`) to see if they successfully deter unilateral deviation via `ExtendsToNash`.
* **Drops Fake Roots:** If no pure strategy can maintain the Nash condition, the equilibrium is mathematically proven to be a false positive and is discarded.
* **Duplicate Filter:** Added a strict equality check to prevent mathematically identical profiles from populating the final solutions list.

**2. Python Tests (`tests/test_nash.py`)**
* Removed the `pytest.mark.xfail` wrappers from the Trust Game tests.
* Updated `game1` (standard trust game) to correctly expect the `[[0.5, 0.5]]` centroid fallback equilibrium.
* Updated `game2` (`unique_NE_variant=True`) to correctly expect the single mathematically valid pure equilibrium: `[[[[1, 0]], [[0, 1]]]]`.

### Testing
* Ran the full test suite locally. **All 105 tests in `test_nash.py` now pass successfully.**

### DIFF
--- a/src/solvers/enumpoly/efgpoly.cc
+++ b/src/solvers/enumpoly/efgpoly.cc
@@ -184,7 +184,69 @@ std::list<MixedBehaviorProfile<double>> SolveSupport(const BehaviorSupportProfil
         data.m_support.ToMixedBehaviorProfile(ToSequenceProbs(data, root)));
     if (ExtendsToNash(sol, BehaviorSupportProfile(sol.GetGame()),
                       BehaviorSupportProfile(sol.GetGame()))) {
-      solutions.push_back(sol);
+
+      Gambit::MixedBehaviorProfile<double> full_sol(sol.GetGame());
+
+      // Copy valid probabilities to the dense full_sol array
+      for (const auto& player : full_sol.GetGame()->GetPlayers()) {
+        for (const auto& iset : player->GetInfosets()) {
+          for (const auto& act : iset->GetActions()) {
+            full_sol[act] = sol.GetActionProb(act);
+          }
+        }
+      }
+
+      bool valid_eq = true;
+      for (const auto& player : full_sol.GetGame()->GetPlayers()) {
+        for (const auto& iset : player->GetInfosets()) {
+          double sum = 0.0;
+          for (const auto& act : iset->GetActions()) {
+            sum += static_cast<double>(full_sol.GetActionProb(act));
+          }
+
+          if (sum < 0.5) {
+            bool found_pure = false;
+            for (const auto& act : iset->GetActions()) {
+              full_sol[act] = 1.0;
+              if (ExtendsToNash(full_sol, BehaviorSupportProfile(full_sol.GetGame()),
+                                BehaviorSupportProfile(full_sol.GetGame()))) {
+                found_pure = true;
+                break;
+              }
+              full_sol[act] = 0.0;
+            }
+            if (!found_pure) {
+              valid_eq = false;
+              break;
+            }
+          }
+        }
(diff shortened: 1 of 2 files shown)


satyam102006:
Hi @tturocy , I completely understand why my PR was close I apologize for jumping the gun before discussing it! I am used to a very fast-paced workflow in other open-source projects and was just eager to dive into the Gambit codebase.

I have formally sent my CV and Cover Letter to your email as per the GSoC Contributor Guidance. Once you have had a chance to review my profile, I would love to discuss the proper approach to solving this issue (or any other issue you recommend) to officially evaluate my technical fit. Looking forward to hearing from you

PR #782, opened 2026-02-21
ENH: Warn when use_strategic has no effect

### LINKED ISSUE #720: Alternatives to the use_strategic argument for solvers

_No response_

In solvers like `lp_solve` and `lcp_solve`, at the moment `use_strategic` defaults to `False`.

When the game does not have a tree representation, we must use the strategic representation, and so `use_strategic` has no effect. 

It is not ideal that when a user has a non-tree game and passes in `use_strategic` as `False` they get no feedback to say that we must use the strategic form in this case. This is not a bug, the docs are clear, it just does not seem idea.

If the default was `True` (not suggesting that we should do this), then at least one could give a warning if one ever tries to set it as `False` for a non-tree, which is more consistent in some sense with always being able to use the strategic representation but not always having a tree. I note that having a `use_tree` argument that defaults to `True`, has the symmetric issue of when a user tries to set it explicitly to `True` when calling on a non-tree -- it makes not sense for them to include that argument in that case, but there is also no way to give feedback since that is the default.

Maybe the current setup is fine, but this did all lead me to wonder if we simply want different functions for the solvers on different representations, though I know that would move us away from the approach taken in many modern scientific computing packages.

Shivika-5236:
Closes #720 

This PR adds a warning when use_strategic=False is explicitly passed to solver functions for games that do not have a tree representation.
Right now, if the game is a normal-form (table) game, the solver must use the strategic representation. So even if a user passes use_strategic=False, it has no effect — but they get no feedback about it.
This PR introduces a decorator warn_on_explicit_use_strategic_false in util.py.

The decorator:

Checks if use_strategic=False was explicitly passed
Checks if the game has no tree representation

Emits a UserWarning saying:
"Game has no tree representation; using strategic form."

This keeps the current behavior unchanged. It just makes it clearer for users when their argument cannot actually be respected.

The decorator is applied to:
lcp_solve
lp_solve
enumpoly_solve
logit_solve
logit_solve_branch
logit_solve_lambda

In addition, this PR includes:

Tests to verify the warning behavior for non-tree games.
Tests to ensure no warnings are raised for tree games.
Tests for MixedBehaviorProfile.as_strategy:
Payoff consistency between behavior and strategy profiles.
Manual probability checks on a small sequential example.
No existing functionality is changed. Only warnings and tests are added.

How to review this PR

You can review this PR in three parts:
1. Decorator implementation (util.py)
Check the logic in warn_on_explicit_use_strategic_false.
Make sure:
The warning only triggers when use_strategic=False is explicitly passed.
It correctly checks game.is_tree.
It does not interfere with normal solver execution.

2. Solver updates
Confirm the decorator is correctly applied to all relevant solver functions.
Ensure there are no unintended side effects.

3. Tests
Run the test suite.
Focus on:
test_use_strategic_warning.py
test_behav_as_strategy.py

Verify:
Warnings appear only when expected.
No warnings for tree games.
Solver outputs remain unchanged.

### DIFF
--- a/tests/test_behav_as_strategy.py
+++ b/tests/test_behav_as_strategy.py
@@ -3,10 +3,10 @@ import pygambit as gbt
 
 @pytest.fixture
 def small_efg():
-    """Fixture providing a minimal 2-player perfect-info sequential game."""
+    
     g = gbt.Game.new_tree(players=["Alice", "Bob"], title="Small Sequential")
     g.append_move(g.root, "Alice", ["L", "R"])
-    # Note: append_move to individual node gives Bob two distinct information sets
+    
     g.append_move(g.root.children[0], "Bob", ["u", "d"])
     g.append_move(g.root.children[1], "Bob", ["u", "d"])
     
@@ -17,14 +17,11 @@ def small_efg():
     return g
 
 def test_as_strategy_payoff_consistency(small_efg):
-    """
-    Test that the expected payoffs match between computing with the original 
-    behavior profile and the converted mixed strategy profile.
-    """
+
     g = small_efg
     profile = g.mixed_behavior_profile(rational=True)
     
-    # Assign some arbitrary probabilities
+
     profile["L"] = gbt.Rational(1, 4)
     profile["R"] = gbt.Rational(3, 4)
     profile[g.players["Bob"].infosets[0]] = [gbt.Rational(1, 2), gbt.Rational(1, 2)]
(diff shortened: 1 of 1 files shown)


tturocy:
We are not able to review unsolicited pull requests that have not been previously discussed with a developer.

If you are interested in participating in GSoC 2026, please read https://www.gambit-project.org/gsoc_2026/ and specifically "Contributor Guidance".  With that information we can assess whether you are likely a good candidate to work on Gambit and provide with suggestions and guidance on initial issues to try out.

PR #816, opened 2026-03-19
Clarify definitions of behaviour profile values for unreached information sets

### LINKED ISSUE #446: ENH: action values for zero-probability infosets should be undefined
Currently, the action value (stored in `map_actionsValues`) for actions at zero-probability information sets are zero. This enhancement would instead have these recorded as undefined. This may require changing the type of values of the map which is currently, via a template, either a double of `Gambit.Rational`, or just not populating it for the respective actions at zero-prob infosets.

tturocy:
This clarifies how quantities related to mixed behaviours profiles are defined for information sets which are not reached.

* Beliefs, action values, infoset values, and node values are not well-defined in this case. Therefore, these functions now return a `std::optional` in C++, with `std::nullopt` when not defined. The Python equivalents now return `None`

* Agent regret concepts (action regret, infoset regret, max agent regret) are well-defined despite information sets not being reachable: from the definition of the (multi)agent form of a game, these are zero.

* Along similar lines to the above, in the agent Lyapunov value, the contribution to the value of unreached information sets is set to zero.

Closes #446.


src/pygambit/behavmixed.pxi
```diff
@@ -704,6 +716,10 @@ class MixedBehaviorProfile:
             If `action` is a string and no action in the game has that label.
         ValueError
             If `action` resolves to an action that belongs to the chance player
+
+        See also
```
rahulsavani: I know this is a bit OCD, but we mostly have "See Also" and a small number of cases of "See also" (lower case "a"). I'd prefer to be consistent.

It wasn't clear to me why we point to `infoset_prob` here under `action_value` but not under `infoset_value`, perhaps just an omission?

tturocy: Entirely because I did these on different days and wasn't consistent with myself.

Issue #394, opened 2023-10-26
MAP: Library of standard games/game generators


tturocy:
## Overview

A nice feature of some packages and libraries - such as R - is that they provide built-in access to "standard" example datasets.  We do have a (admittedly not well curated) set of examples which are mentioned in the documentation, and which are kept in `contrib/games`.

The proposal is that we should package these standard games at least in `pygambit`, and make it easy to load them, perhaps by a submodule that allows them to be loaded by name (e.g. `pygambit.examples.load('myerson_poker')`).  In this collection it would be lovely to include all of the extensive and strategic games from prominent game theory textbooks or important articles.

In addition, it would be useful to develop standard codes for building certain example families/classes of games - for example, Kuhn poker with varying deck sizes.

These would be quite useful to users as examples or starting points.  In addition, they would be excellent to integrate within the test suite - as they would be natural games to use for testing purposes.  (In fact, we might consider that as far as the test suite goes, we should either use these "standard" games or, if we need bespoke games for specific tests, either to build them using the API calls or embed the game file representation in the test suite module, rather than having separate .efg/.nfg files.)

Some discussion should be had on what's the best way to organise the calling patterns for this, to accommodate both having standard games serialised explicitly (whether in `.efg`, `.nfg`., `.agg`, or other formats), and functions which generate games based on payoff or structural parameters.

PR #827, opened 2026-03-25
Add games from Gilboa (1997) and Jakobsen et al. (2016) to catalog

d-kad:
### Issues closed 
- Contributes to #394

### Description of the changes
This PR adds five games to the catalog that share a common feature: the absence of a public clock.

- `gilboa1997/fig2.efg` — Two-selves reformulation of the absent-minded driver from Gilboa (1997)
- `jakobsen2016/fig1a.efg` — Not exactly timeable coin-toss guessing game from Jakobsen et al. (2016)
- `jakobsen2016/fig1b.efg` — Exactly timeable variant of the above
- `jakobsen2016/fig1c.efg` — Not exactly timeable variant where the second bet is conditional
- `jakobsen2016/fig3.efg` — Four-player onion routing game, not exactly timeable

Bibliography entries for Gil97 and JakSorCon16 have been added.

Makefile.am has been updated.


edwardchalstrey1 (commented):
Looks good to me @d-kad I see you already noticed we changed yesterday how the linking works to use the full https://gambitproject.readthedocs.io/en/latest/biblio.html url, which I was about to email you about. I've just merged in the latest catalog infrastructure updates to the master branch, I think you should be able to just merge the master branch  into your branch without any difficulty other than fixing the bibliography conflict.

d-kad:
Thanks for a quick review @edwardchalstrey1! 
I merged master, resolved the bibliography conflict, and fixed a Unicode issue in one of the descriptions. 
Ready for another look.

tturocy:
Also, I would like to merge this after we release 16.6.  This is only because we have a very small catalog at the moment and these are exotic examples, so it would be a bit odd I think to be having them while a lot more "obvious" examples are not yet included!

PR #828, opened 2026-03-25
Adjust packaging of catalog games into Python build.

### LINKED ISSUE #823: [PyGambit]: Configuration of catalog
The configuration of how we're packaging the catalog in `pygambit` doesn't look quite right.  I am getting the following warning for each of the directories in the catalog.

```
/private/var/folders/33/szs2dgz97j7dhgx5zdfjfzs40000gn/T/pip-build-env-ijz85bne/overlay/lib/python3.13/site-packages/setuptools/command/build_py.py:215: _Warning: Package 'pygambit.catalog.myerson1991' is absent from the `packages` configuration.
  !!

          ********************************************************************************
          ############################
          # Package would be ignored #
          ############################
          Python recognizes 'pygambit.catalog.myerson1991' as an importable package[^1],
          but it is absent from setuptools' `packages` configuration.

          This leads to an ambiguous overall configuration. If you want to distribute this
          package, please make sure that 'pygambit.catalog.myerson1991' is explicitly added
          to the `packages` configuration field.

          Alternatively, you can also rely on setuptools' discovery methods
          (for example by using `find_namespace_packages(...)`/`find_namespace:`
          instead of `find_packages(...)`/`find:`).

          You can read more about "package discovery" on setuptools documentation page:

          - https://setuptools.pypa.io/en/latest/userguide/package_discovery.html

          If you don't want 'pygambit.catalog.myerson1991' to be distributed and are
          already explicitly excluding 'pygambit.catalog.myerson1991' via
          `find_namespace_packages(...)/find_namespace` or `find_packages(...)/find`,
          you can try to use `exclude_package_data`, or `include-package-data=False` in
          combination with a more fine grained `package-data` configuration.

          You can read more about "package data files" on setuptools documentation page:

          - https://setuptools.pypa.io/en/latest/userguide/datafiles.html


          [^1]: For Python, any directory (with suitable naming) can be imported,
                even if it does not contain any `.py` files.
                On the other hand, currently there is no concept of package data
                directory, all directories are treated like packages.
          ********************************************************************************

```

I am thinking the problem here is that we have the Python code in `catalog/__init__py`.  That should be in perhaps `pygambit/catalog.py` instead.  Then we should copy over the catalog to a different directory (maybe `catalog_data` or `catalog_games`?  I am thinking it's the `__init__.py` that is the core of the problem, as that's what makes it look like a package.

It would also be much cleaner to keep all of the Python code under `src/pygambit` rather than having some of it copied over at build time.

tturocy:
This makes a few adjustments to how we include the catalog games in the Python build.

Support for 'package data' in Python isn't great currently, and further we have a desire to keep the catalog
games elsewhere
in our tree and not embedded inside the
Python source.

Our solution is this:
* We have a custom build class, which is responsible for copying the contents of the catalog directory from where it lives into the build directory.
* We put this in "catalog_data" to avoid clashes with e.g. just having a "catalog" source file or module.

The extra build step is moderately annoying but
straightforward.
However, it avoids e.g. `pip` thinking that the
catalog data is Python code and complaining
about the package not being listed.

Closes #823.

### DIFF
--- a/doc/developer.catalog.rst
+++ b/doc/developer.catalog.rst
@@ -8,18 +8,6 @@ To do so, you will need to have the `gambit` GitHub repo cloned and be able to s
 you may wish to first review the :ref:`contributor guidelines <contributing>`.
 You'll also need to have a developer install of `pygambit` available in your Python environment, see :ref:`build-python`.
 
-The catalog module
-------------------
-
-Although the ``catalog`` directory is located at the project root outside of ``src/pygambit/``, it is installed and bundled as the ``pygambit.catalog`` subpackage.
-
-This is handled by the package build configuration in ``pyproject.toml`` under ``[tool.setuptools]``:
-
-- The ``package-dir`` mapping instructs ``setuptools`` to source the ``pygambit.catalog`` subpackage from the physical ``catalog`` directory.
-- The ``package-data`` configuration ensures all non-Python data files (like ``.efg`` and ``.nfg`` files) inside the catalog are correctly bundled during installation.
-
-As a developer, this means you will need to reinstall the package (e.g., passing ``pip install .``) for any new game files or internal catalog changes to be reflected in the ``pygambit`` module.
-
 Add new game files
 ------------------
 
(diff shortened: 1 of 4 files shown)


tturocy:
@edwardchalstrey1 Here's a candidate solution to the issue.  I think it's coherent and it's resolved the warning for me.  It is slightly flddly but that's the cost of us wanting to be able to pull the catalog data from elsewhere (which we will definitely want to do, as someday it will possibly be its own repo which we e.g. refer to by a submodule!)

Question: Do we need to memorialise this anywhere in the developer documentation?

edwardchalstrey1 (commented):
Great, this solution works for me!

Regarding docs, the process for updating the catalog is unchanged, however I did add the header ["The catalog module"](https://gambitproject.readthedocs.io/en/latest/developer.catalog.html) before our discussion, which is now out of date. This could either be updated or simply removed, depending how useful you think it is for devs on the project to read vs just inspecting the codebase.

PR #835, opened 2026-04-02
Fix infoset realisation probability for absent-minded information sets

### LINKED ISSUE #826: [Bug]: Information set realisation probability for games with absent-mindedness
The current implementation of `MixedBehaviorProfile` computes information set realisation probability by summing the realisation probabilities of its member nodes.

This assumes non-absent-mindedness.  For an information set that exhibits absent-mindedness, this "double counts" multiple visits to the information set (and in principle would allow an information set to be visited with probability greater than one!)

We propose that the correct extension of information set realisation probabilities to the case of absent-minded information sets is that it would be the probability the information set is visited *at least once*.  Because node realisation probabilities are retained (and we have the ability to check whether an information set exhibits absent-mindedness) it is possible to compute the realisation structure in more detail if desired.

Currently the information set realisation probability is computed on demand.  It would likely make sense to change this to be a cached value for each information set.

d-kad:
### Description

This PR fixes information set realisation probability for games with absent-mindedness. The previous implementation summed node realisation probabilities across all members, double-counting it for the nodes along the same path, with probabilities summing above 1 in this case.

The corrected infoset realisation probability is the probability the information set is visited at least once. The implementation identifies reentry nodes (members whose path from the root passes through the same information set) and excludes them from the sum.

**C++ changes:**
- Add `m_absentMindedReentries` cache in `GameTreeRep`, populated inside `BuildOwnPriorActions`
- Add `GetAbsentMindedReentries`
- Add `m_infosetProbs` cache, populated inside `ComputeRealizationProbs`

**Test changes:**
- Add `test_absent_minded_infoset_prob` covering `noPR-AM-driver-one-player.efg`, `noPR-action-AM.efg`, and `noPR-action-AM-three-chain.efg` with uniform probability distributions over actions

Closes #826


tturocy:
The only thing that is giving me pause here is whether the need to run the unreachable nodes check and then the way the absent-minded nodes check is being done is too inefficient.

Wouldn't it be simpler to do the sums and then subtract off the probabilities from the re-entries?  That way if the set of re-entries is empty that would be fast.

tturocy:
We've realised with further discussion that there are some additional adjacent considerations to defining some other quantities that should be handled as part of these changes (as they interact with the definition of information set realisation probability) - see discussion to be added to #826

tturocy:
@d-kad I am having difficulty correctly resolving the merge conflict in the tests - as you've been working on these I would hope you would be able to see the correct solution more quickly than I!

tturocy (approved):

PR #842, opened 2026-04-17
User flow on launching GUI without a game loaded

### LINKED ISSUE #80: ENH: Re-use graphical interface window from trivial game
At the moment, the sequence (1) launch Gambit and then (2) open a game file results in two Gambit windows, the first containing a trivial game and the second the opened game.

Consider re-using a window containing a pristine, unedited "trivial" game -- if one is present -- to load a new game file, to minimise window clutter.

tturocy:
This is WIP towards a solution to #80.

Rather than re-using a frame as suggested by #80, this proposes a bit more modern style, whereby if a game is not loaded on the command-line (or otherwise at launch), there's a "welcome" frame that offers the ability to load a game, or to create a new one.  This "welcome" frame is then dismissed after a game is loaded.  So, this has roughly the same effect as intended by #80, although the mechanism is slightly different.

I would like to get views on:
* Does this work well / look decent on other platforms (I've only looked at MacOS);
* Any suggestions for improvements, or things that don't quite work right (try to break it!)


tturocy:
I have also added to this now a new dialog that prompts for dimensions of a normal form game before creating it - I think it make sense to bundle this here as it's all under "ease of startup/creating a new game".

PR #856, opened 2026-04-22
Better UX for initial game sizing, focusing on selected nodes

### LINKED ISSUE #569: Improve behaviour of making nodes visible / scrolling the extensive game display

Gambit GUI

We have a feature in the graphical interface (hitting the space bar) which scrolls the window to make the currently-selected node visible.  This usually does manage to make the selected node visible - but it's not necessarily a slick or modern result (for example the selected node isn't centred at all even when it might make sense to be able to centre it).

Relatedly, when a game is first displayed, the window is not scrolled to display the root node of the game and/or the zoom level of the game is not set appropriately.  So for even a rather moderately-sized game, the initial display window is blank because the tree is actually down and to the right of the part of the canvas that is being displayed.

Given the interconnections between these two - basically they both boil down to trying to make working with medium-sized games a bit more smooth - it makes sense to think of these as part of the same project to be solved.

tturocy:
This  implements several related minor but useful UX improvements for game trees:

* Large trees are initially zoomed to a reasonable size (clamped not to be too small)
* The root node is ensured to be visible when the game is first displayed
* Pressing the space bar to ensure the selected node is visible now works correctly (by fixing the event type for keypresses)
* The initial size of the frame for the game window is now adaptive relative to display size.

Closes #569.


edwardchalstrey1:
Screenshot 2026-04-30 at 10 54 42

- I have tested this and the space bar feature works for me.
- The root node for a large game just about visible when opening full screen
- Initial zoom looks good to me, as does game window frame on different screen sizes

edwardchalstrey1 (approved):

PR #859, opened 2026-04-27
Fix regression in action graph games

tturocy:
This fixes a regression in action graph games, that was causing a segmentation fault due to not initialising the strategy index correctly after doing so was introduced in a refactor a few months ago.

I have tested this with `2x2.agg` and the previous behaviour has been restored.

While we are not planning to do anything major with action graph games for 17.0, this would be a good opportunity for us to add some entries into the test suite just to guard against regressions on the limited set of AGG files we currently have.


rahulsavani (commented):
I can confirm that the fix seems to work fine. 

I will now add tests and then ask @tturocy to review those.....

tturocy:
I've now added support for outcomes in (Bayesian) AGGs in Python, which indirectly solves the `.to_arrays()` problem (and also #871)

tturocy:
@rahulsavani In view of the discussion in #879, would it make sense only to run tests in floating-point for the purposes of considering the regression identified in this PR as "fixed"?  The broader issue of the semantics of numerical data types in (B)AGGs has always been there (we have just never thought carefully about it), so the fix in this PR would at least restore us to the world before the bug was introduced.

The plan in #879 would then point towards further tests to add in due course.

rahulsavani:
> @rahulsavani In view of the discussion in #879, would it make sense only to run tests in floating-point for the purposes of considering the regression identified in this PR as "fixed"? The broader issue of the semantics of numerical data types in (B)AGGs has always been there (we have just never thought carefully about it), so the fix in this PR would at least restore us to the world before the bug was introduced.
 
> The plan in #879 would then point towards further tests to add in due course.

Absolutely, that's what I had in mind and planned to discuss with you. I will update the tests accordingly asap.

tturocy:
To capture from an offline conversation: There is nothing special about `2x2.agg` and it would be a good idea at a minimum to trim the decimals to fit within standard `double` just to eliminate one source of potential cross-system variation.  Having "nice" payoffs where we also know the answer easily is useful.

With this PR we're much more concerned with having some tripwires for unintentional regressions than having a comprehensive suite - that's a bigger and separate concern.

rahulsavani:
@tturocy This is ready for review and hopefully merging. `Bayesian-Coffee-3-2-2-3.bagg` is too big for `enumpoly` but we have tests with `enumpure` and `logit`, which is probably fine for now.

PR #869, opened 2026-05-07
Use catalog games for tests

tturocy:
This removes some games which were still in `test_games`, but now have been canonicalised in the catalog.


rahulsavani (commented):
This looks good to me. Bye bye e01.efg and e02.efg.

BTW, while the changes to the test suite look good as part of this review I "grep"ed e01.efg and e02.efg and found references to them in the docs in tools (in tools.enumpoly.rst for e01.efg and in {liap,lcp,simpdiv,enumpure}.rst for e02.efg). Perhaps you could just fix that in this PR.

tturocy:
Good point.  We don't have a great way to reference those (as paths would depend on the `cwd` of the shell) but I've done something sensible to make it clear they're catalog games.  We'll likely be re-drafting this documentation with the planned reworking of the command-line tools anyway so this should do for present purposes.

Issue #760, opened 2026-02-05
[Catalog] 3rd party game libraries inclusion


edwardchalstrey1:
### This feature request is for...

PyGambit

### Requested enahancement or feature

- We want Gamut to be a subset of the catalog
- Can also load OpenSpiel/SageMath games with the same names as part of the catalog?
- How can we keep external libraries optional dependencies?
- Code could be like: `pygambit.catalog.load("openspiel_tinyhanabi")` which fails if OpenSpiel not installed OR better still; just have the EFGs from OpenSpiel in Gambit Catalog.
    - Note: We may need to do a PR to OpenSpiel to ensure the EFG output works for all games

edwardchalstrey1:
# OpenSpiel

Update the Gambit catalog such that we can load games from [OpenSpiel](https://openspiel.readthedocs.io/en/latest/games.html) via one of [EFG exporter in python](https://github.com/google-deepmind/open_spiel/blob/master/open_spiel/python/algorithms/gambit.py), 
[Independent EFG exporter in C++](https://github.com/google-deepmind/open_spiel/blob/master/open_spiel/game_transforms/efg_writer.cc) or [NFG exporter written in C++](https://github.com/google-deepmind/open_spiel/blob/master/open_spiel/algorithms/nfg_writer.cc), noting that some games will not correctly generate a Gambit EFG or NFG and we should ignore those games for now.

You should be able to load these games via the catalog module like:
```{python}
pygambit.catalog.load("open_spiel/Atari")
```

TODO:
- [x] Update script that generated open_spiel games
- [x] Single dev dependency list
- [x] Also find a way to add descriptions to open_spiel games

PR #878, opened 2026-05-21
Add games from OpenSpiel to catalog

edwardchalstrey1:
### Issues closed by this PR

Described in #760 but this only does OpenSpiel and only games that currently work

This PR adds:
- 13 new games to the catalog, derived from OpenSpiel
- A new script: `build_support/catalog/generate_openspiel.py` - the intention here is to run this whenever new games are added to OpenSpiel that we want included in the Gambit catalog
    - Currently most of the games in the OpenSpiel library cannot be exported to Gambit. Only **13 / 122** are included here. Mostly small strategic games that we likely already have in `contrib` (a 14th one "First-price Sealed-Bid Auction" was too big so I didn't include it).
    - If we re-run this script after the [exporters](https://www.gambit-project.org/gsoc_2026/#idea-8-efgnfg-exporters-for-deepminds-openspiel) are updated, we could get a much larger number of games. It would probably be best to do this once the catalog is more mature
    - The approach currently is to attempt for each game in `pyspiel.registered_games()` to first create an NFG, then if that fails try EFG, then give up and print the name of the game that couldn't be exported
- Also consolidates a single "dev" dependency list in `pyproject.toml` which includes `open_spiel` where previously there were separate "doc" and "test" lists

### How to review this PR

@tturocy @rahulsavani  take a look at the catalog page: https://gambitproject--878.org.readthedocs.build/en/878/catalog.html#catalog and from "Coordinated Matching Pennies" downwards you'll see the games added from OpenSpiel. Questions:

1. Do you think any of these 13 games are worth including in the catalog now, or are they all things that we'll likely have from other sources anyway?
    - One exception might be Tiny Hanabi (I can fiddle a bit more with the DrawTree settings on this one if so)
2. Regardless of the answer to 1, do you think my approach with the `generate_openspiel` script (see above) is the best approach, as oppose to making a call to the OpenSpiel package when loading a game from there via PyGambit? I think this way gives us a lot of control over what goes in the catalog, whilst potentially giving us a lot of new games to add once the exporters work properly.


edwardchalstrey1:
May be some extra games possible due to some games size being a factor. Some of the OpenSpiel games have parameters that affect the size.

Alternative approach: load from the pyspiel game library directly but raise an exception if it's not installed. Don't have it as a dependency

```{python}
def load_openspiel():
    g = ops_game.to_efg()
    return gbt.read_efg(g)
```

**Function that takes a func that has the same args as `pyspiel.load_game` and then tries to do the nfg or efg generation as in the generate_openspiel script and raises not installed exceptions or any other exceptions**

edwardchalstrey1:
Working on a new branch `opespiel-catalog-alt`

PR #883, opened 2026-05-24
Fix core dump on degenerate games by returning trivial zero-dimensional profile

AndresFerCervell:
**PR Title:** Fix core dump in logit solver for degenerate games (zero decisions)

### Issues closed by this PR

None. (Standalone bug fix discovered during the investigation of #492).

This PR fixes a core dump that occurred when running the QRE logit solver on degenerate games (games where no player has any decision nodes, such as `chance_root_5_moves_no_nonterm_player_nodes.efg`).

- Implemented an early exit defensive shield in `src/solvers/logit/[e/n]fglogit.cc` to detect games with a 0-dimension strategy space.
- By analogy to games with zero payoffs, the solver now cleanly constructs and returns the trivial equilibrium profile (with `lambda = 0`) instead of attempting to build an empty Jacobian matrix or throwing a runtime error.

### How to review this PR

- Review the early exit logic added to `src/tools/logit/logit.cc` after the game file is read.
- Verify the fix by running `./gambit-logit tests/test_games/chance_root_5_moves_no_nonterm_player_nodes.efg`. It should now cleanly print `NE` and exit with code 0 instead of throwing a segmentation fault.
- Run the standard test suite (`make check`) to ensure `.nfg` games are processed correctly and do not trigger `Undefined operation on game` errors.

### DIFF
--- a/src/solvers/logit/efglogit.cc
+++ b/src/solvers/logit/efglogit.cc
@@ -313,6 +313,9 @@ LogitBehaviorSolve(const LogitQREMixedBehaviorProfile &p_start, double p_regret,
                    double p_firstStep, double p_maxAccel,
                    MixedBehaviorObserverFunctionType p_observer)
 {
+  if (p_start.size() == 0) {
+    return {p_start};
+  }
   PathTracer tracer;
   tracer.SetMaxDecel(p_maxAccel);
   tracer.SetStepsize(p_firstStep);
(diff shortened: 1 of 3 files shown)


tturocy:
@AndresFerCervell The place to put this guard is in the algorithm itself, rather than the C++ wrapper.  The reason is I am going to assume we get the same error when we try this in Python.

We don't have a C++ based test suite (so `make check` is a no-op) - we primarily test using the Python wrapper using `pytest`.  So that would be the place to add a failing test that's fixed by this.

For running the check - notice that the C++ functions in `src/solvers/logit` all take profiles as arguments.  These profiles are automatically the correct shape for the underlying game.  So you can simply check whether those profiles are zero-length without having manually to compute numbers of actions or strategies!  So `src/solvers/logit/[en]fglogit.cc` is the place to look for inserting this check.

tturocy:
@all-contributors please add @AndresFerCervell for code

PR #889, opened 2026-05-26
Improve catalog update process and script, including tests

### LINKED ISSUE #885: [Catalog]: Add a DrawTree settings config
### Requested documentation improvement

There is currently a function called [catalog_draw_tree_settings](https://github.com/gambitproject/gambit/blob/00c2d404df0b2dd0a7e69963902924ebe5918c3b/build_support/catalog/update.py#L14) which modifies the default DrawTree settings for Gambit catalog games. This likely won't scale well - @rahulsavani has suggested creating a config file which can be more easily updated as the catalog scales to a large number of games.

Linked to #886 

Todo:
- [x] Come up with a suitable yaml format
- [x] Update `update.py` to use the new config
- [x] Update docs

### LINKED ISSUE #886: [Catalog]: Use curated EF files for catalog visuals by default
### Requested documentation improvement

We should update the catalog such that extensive form games can have an included EF file, which supersedes the EFG for use in DrawTree generated visuals.

Related to #885 - these can probably be part of the same PR.

Todo:
- ~~Commit 1 or more EF files for existing catalog games~~
- [x] Re-write the catalog update script to use the EF for generating images if there is one, then use the EFG to generate an EF via the Gambit layout if not

### LINKED ISSUE #887: [Catalog]: Allow multiple images of the same game
### Requested documentation improvement

For some games in the catalog, we may want to include several different variants of the drawn game on the catalog page. For example, `vonstengelforges2008/fig6`:

Image

Image

### Pre-requisites

- #885 
- #886 

### To do

- [x] The catalog update script should detect if there are multiple EF files, or multiple config settings for a particular game
- [x] You should be able to scroll left/right or have some buttons to click through variants

### LINKED ISSUE #888: Simplify how Makefile.am is updated in catalog update script

Other

- [x] Create another file to include into Makefile.am which defines a variable listing all the game files, rather than trying heuristically to parse Makefile.am
- [x] Can then tidy the [updating the catalog](https://gambitproject.readthedocs.io/en/latest/developer.catalog.html) page to not have references to `EXTRA_DIST` etc
- [x] Make it clearer in that docs page in step 3 that locally you should do [make html](https://gambitproject.readthedocs.io/en/latest/developer.contributing.html#editing-this-documentation) after [updating the catalog] (including `pip install .`, then to inspect the generated catalog page (which can also just be done on PR build success - step 4)
- [x] Can we ensure that doing `pip install -e .` or similar works for catalog updates since the catalog module is pure python and the files are in the package data already:
```
[tool.setuptools.package-data]
pygambit = ["catalog_data/**/*"]
```

edwardchalstrey1:
### Issues closed by this PR

- Closes #888
- Closes #885 
- Closes #886 
- Closes #887 

- This PR has several updates to the catalog update process, both in terms of docs and code, see the linked issues above.
- Also updates the catalog module to ensure that non EFG/NFG files are not picked up by mistake - something @rahulsavani noticed
- Adds help to the update script so you can do `python build_support/catalog/update.py --help`
- Adds tests for `build_support/catalog/update.py` covering the functions handling DrawTree settings, updating the variables passed to the `Makefile.am` and correctly generating the catalog RST doc
- Also updates DrawTree to 0.9.1 which includes fixes to EF->EFG generation

### How to review this PR

- Review the developer docs page on how to update the catalog: https://gambitproject--889.org.readthedocs.build/en/889/developer.catalog.html
- And the catalog page itself: https://gambitproject--889.org.readthedocs.build/en/889/catalog.html


doc/developer.catalog.rst
```diff
@@ -13,10 +13,13 @@ Currently supported representations are:
 - `.efg` for extensive form games
 - `.nfg` for normal form games
 
-.. note::
+.. important::
 
    When updating the catalog, changes can be viewed by inspecting the documentation build generated by a pull request.
-   To test changes locally, you'll need to have a developer install of `pygambit` available in your Python environment, see :ref:`build-python`.
+   To test changes locally, you'll need to have an editable developer install of `pygambit` available in your Python environment.
```
tturocy: You don't need to have an editable install, but it can be a useful technique if you are working only with the catalog.  This needs to be written a bit more carefully though because an editable install doesn't work in general if one is making changes to the C++ part of the code.

This edit might be more confusing than helpful to a less-experienced dev.

edwardchalstrey1: Fair enough, I can just make that explicit here - the idea is yes in general developer install would not be editable, but when you're making catalog edits this will definitely save a lot of time

edwardchalstrey1:
@rahulsavani If you have any EF layout variants for the `vonstengel2022` or `vonstengelforges2008` games we can include them in this PR since there is now the option for custom EF layouts and to include multiple for a given game.

edwardchalstrey1:
> @rahulsavani If you have any EF layout variants for the `vonstengel2022` or `vonstengelforges2008` games we can include them in this PR since there is now the option for custom EF layouts and to include multiple for a given game.

Example of how this looks adding 2 EFs for one game when I tested it locally:

Screenshot 2026-05-29 at 09 37 43

rahulsavani:
> > @rahulsavani If you have any EF layout variants for the `vonstengel2022` or `vonstengelforges2008` games we can include them in this PR since there is now the option for custom EF layouts and to include multiple for a given game.
> 
> Example of how this looks adding 2 EFs for one game when I tested it locally:
> 
> Screenshot 2026-05-29 at 09 37 43

This looks great. I am reviewing the PR now.

One immediate comment: The long list of files does not look great -- can we hide it (within the entry) by default and expand it similarly to the whole entry?

edwardchalstrey1:
> The long list of files does not look great -- can we hide it (within the entry) by default and expand in similarly to the whole entry?

Good idea, have changed that for all the entries and it looks better

rahulsavani (changes requested):
I really like the tabs for alternatives. 

However, at the moment, to get two or more alternatives one needs to add two or more .ef files to the catalog.

I think a typical use case is that one provides one alternative with the other desired one  is that created by`draw_tree` (i.e. what we used to have before allowing custom .efs at  all).

Currently if one provides a  `{slug}__{label}.ef` without a  `{slug}.ef` the former is ignored and one just gets the default `draw_tree` one. I think in this case the default and the variant should be created.

edwardchalstrey1:
> Currently if one provides a `{slug}__{label}.ef` without a `{slug}.ef` the former is ignored and one just gets the default `draw_tree` one. I think in this case the default and the variant should be created.

@rahulsavani Just pushed an updated and that is now the case

edwardchalstrey1:
Just also added a small change to the recently added `gilboa1997/fig1` to prevent the action label overlapping the information set boundary:

In `draw_tree_settings.yml`:
```{yaml}
gilboa1997/fig1:
    action_label_dist: 5.0
```

Screenshot 2026-06-01 at 11 00 36

rahulsavani (approved):
Thanks, this works nicely. I am happy with this for now.

I added two alternative .efs for vonstengelforges2008; they need some minor tweaks, but I will do that later -- let's get this merged in so you can merge in the bibtex stuff.

PR #893, opened 2026-05-28
Bernhard catalog 7 games

rahulsavani:
3 games from von Stengel and Forges (2008), which is all from that paper, and 4 games from von Stengel (2022), which is a small subset from that book. There's a brief comment for each, which we may expand on in due course.


edwardchalstrey1 (approved):
All looks great to me! Let's merge this before #889 and I can update that PR with these new games.

Issue #897, opened 2026-05-28
[DEV]: Remove use of `tinyxml`


tturocy:
We are moving away from the XML-based `.gbt` format - which we have never supported properly anyway - in favour eventually of better formats, especially as our graphical/visual tools evolve.  We no longer need to carry around `tinyxml` as an internal dependency to accomplish what we need to do.

### Tasks

* [ ] `pygambit`: When reading a `.gbt` file use a Python XML-reading package to extract the game.
* [ ] Port graphical interface XML file reading/writing to the wxWidgets XML classes

Issue #896, opened 2026-05-28
[DEV]: Remove uses of `wxSheet` from graphical interface


tturocy:
We will remove the remaining uses of `wxSheet` in favour of `wxGrid`, which has had many improvements since we started using `wxSheet`, and is core wxWidgets.   In some cases, a grid isn't the best solution for UX anyway; in those cases the replacement shouldn't be mechanically done like-for-like but should change the interface to a more appropriate design.

### Tasks

The tasks in this list are ordered roughly in increasing complexity.

* [ ] `dleditmove.cc`
* [ ] `efglogit.cc` / `nfglogit.cc`
* [ ] `efgprofile.cc`/`nfgprofile.cc`
* [ ] `nfgtable.cc`

Issue #894, opened 2026-05-28
[MAP]: Port command-line tools to Python


tturocy:
After discussion, we have agreed that from 17.0, the command-line interfaces for solving games will be written in Python, as opposed to the current C++ wrappers.  This will enable much easier maintenance (by using for example `click` for command-line argument parsing; proper, robust CSV reading; and ease of formatting outputs, including outputs in more machine-readable formats like JSON).

### Tasks

* [x] GUI to use threads for long-running calculations rather than external binaries (so we do not have to worry about Python dependency if we don't want to)
* [ ] Generate Windows .EXE as part of the build/distribution process for binaries.

PR #907, opened 2026-06-01
Draft overview of architecture

tturocy:
This adds a section to the docs giving a brief overview of the conceptual architecture of Gambit.

This is written with an eye to it being accurate for Gambit 17.  Therefore, it ignores the various remaining "third-party" bits of code we are removing and for which issues are already open (see e.g. #897 , #896 ).  It also anticipates the plan to have command-line tools be written in Python rather than C++ (see #894 ).

In drafting this, an observation comes to mind that without the command-line tools, `make` in practice now is entirely about the graphical interface (and so if wxWidgets is not installed/available then `make` does nothing).  The disclaimer that the C++ does not constitute a "library" seems awkward - and perhaps this indicates we should build the core as a C++ library (even if it comes with a disclaimer that as yet the API is not promised to be as stable as `pygambit` at least for 17.x?)

### DIFF
--- a/doc/developer.overview.rst
+++ b/doc/developer.overview.rst
@@ -1,0 +1,43 @@
+.. _architecture-overview:
+
+Overview of Gambit architecture
+===============================
+
+The core of Gambit is a collection of C++ code for the representation and analysis of finite games.
+This code is organised into three directories:
+
+* ``src/core``: Generic functions, structures, and utility code which is not specific to
+  game theory.
+* ``src/games``: Definition of the interface and representation data structures for game objects
+  and related concepts.
+* ``src/solvers``: Implementation of algorithms for computation on games, principally for the
+  computation of (exact or approximate) Nash equilibria.
+
+At present this core is referred to as a "collection" rather than a "library" as
+this is not (yet) packaged by the build system as a reusable C++ library with a documented,
+stable interface.
+
+Gambit is written to the C++17 standard.  Because Gambit is cross-platform, C++ library
+dependencies are avoided to facilitate development and packaging.
+
+The principal way in which code is written to use Gambit is via the Python package
+``pygambit``.  In particular, testing is done via the Python test suite and ``pytest``,
+which therefore (indirectly) tests the C++ data structures and algorithms.
+Running ``make check`` has no effect (there are no tests defined in that build system).
+
+From Gambit 17, ``pygambit`` is also responsible for providing command-line interfaces to
+equilibrium computation and other facilities; previously the command-line wrappers were
+implemented in C++.
+
+Gambit's core is also incorporated into a graphical user interface, also written
+in C++ using wxWidgets (see ``src/gui``).  This is built by default by ``make``.
+
+This hybrid architecture aims to maintain some flexibility for future development, by
+
+1.  Although in practice we expect ``pygambit`` will be the most common path for writing
+    code using Gambit, the C++ implementations are kept separate from the Python wrapper
+    so they can continue to be used directly in C++
+2.  Likewise, the graphical interface layer is kept cleanly separate from the core.
+    The developers anticipate that graphical-based interfaces will evolve rapidly in
+    the near future, both in terms of UX and implementation details (different
+    libraries, languages, browser-based interfaces).
--- a/doc/developer.overview.rst
+++ b/doc/developer.overview.rst
@@ -39,4 +39,3 @@ This hybrid architecture aims to maintain some flexibility for future developmen
     so they can continue to be used directly in C++
 2.  Likewise, the graphical interface layer is kept cleanly separate from the core.
     Development of graphical-based interfaces (different libraries, languages, browser-based interfaces) can be carried out without touching the core code.
-    libraries, languages, browser-based interfaces).
--- a/doc/developer.overview.rst
+++ b/doc/developer.overview.rst
@@ -3,7 +3,7 @@
 Overview of Gambit architecture
 ===============================
 
-The core of Gambit is a collection of C++ code for the representation and analysis of finite games.
+The core of Gambit is a library of C++ code for the representation and analysis of finite games.
 This code is organised into three directories:
 
 * ``src/core``: Generic functions, structures, and utility code which is not specific to
@@ -13,10 +13,6 @@ This code is organised into three directories:
 * ``src/solvers``: Implementation of algorithms for computation on games, principally for the
   computation of (exact or approximate) Nash equilibria.
 
-At present this core is referred to as a "collection" rather than a "library" as
-this is not (yet) packaged by the build system as a reusable C++ library with a documented,
-stable interface.
-
 Gambit is written to the C++17 standard.  Because Gambit is cross-platform, C++ library
 dependencies are avoided to facilitate development and packaging.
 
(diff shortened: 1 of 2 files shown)


StephenPasteris (approved):

doc/developer.overview.rst
```diff
@@ -0,0 +1,43 @@
...
+Gambit's core is also incorporated into a graphical user interface, also written
+in C++ using wxWidgets (see ``src/gui``).  This is built by default by ``make``.
+
+This hybrid architecture aims to maintain some flexibility for future development, by
+
+1.  Although in practice we expect ``pygambit`` will be the most common path for writing
+    code using Gambit, the C++ implementations are kept separate from the Python wrapper
+    so they can continue to be used directly in C++
+2.  Likewise, the graphical interface layer is kept cleanly separate from the core.
+    The developers anticipate that graphical-based interfaces will evolve rapidly in
```
edwardchalstrey1: Maybe future-proof this last sentence to avoid needing to remember to update it?
```suggestion
    Development of graphical-based interfaces (different libraries, languages, browser-based interfaces) can be carried out without touching the core code.
```

doc/developer.overview.rst
```diff
@@ -0,0 +1,43 @@
...
+    the near future, both in terms of UX and implementation details (different
```
edwardchalstrey1:
```suggestion

```

doc/developer.overview.rst
```diff
@@ -0,0 +1,43 @@
...
+    libraries, languages, browser-based interfaces).
```
edwardchalstrey1:
```suggestion
```

edwardchalstrey1 (approved):
Looks good to me, I made one suggestion but feel free to reject

doc/developer.overview.rst
edwardchalstrey1: i.e. the key thing I'm removing is "in the near future"

d-kad (approved):

doc/developer.overview.rst
```diff
@@ -0,0 +1,43 @@
...
+stable interface.
+
+Gambit is written to the C++17 standard.  Because Gambit is cross-platform, C++ library
+dependencies are avoided to facilitate development and packaging.
+
+The principal way in which code is written to use Gambit is via the Python package
+``pygambit``.  In particular, testing is done via the Python test suite and ``pytest``,
+which therefore (indirectly) tests the C++ data structures and algorithms.
+Running ``make check`` has no effect (there are no tests defined in that build system).
+
```
AndresFerCervell, line 27: I would add some information about the distribution of games in the repository for testing specific situations while coding in C++. I found it a little confusing at first to find the game files, and to be honest, I still don´t fully understand why they are distributed across different directories (``tests/test_games``, ``contrib/games``, ``catalog``...)

rahulsavani: That is a very fair point. For your info for now:

- The catalog is new and in development. `contrib/games` will disappear as things are moved to the catalog. It is not happening at once because we are adding careful extra data to the entries in the catalog, which will one of the coolest new features of Gambit 17.

- `tests/test_games` will likely persist for games that are not of general interest, unlike those in the catalog

- `src/games` was in your old list but I think you just edited it out, as indeed, it should not contain any games just source code.

tturocy: Yes, this is a good point - as Rahul notes this is something that's in active transition.  The "catalog of games" documentation covers where the endpoint will be, but things are in flight as we curate the existing games (and a whole lot more).

We're writing this with respect to where we'll be with release 17 so mostly this will be covered, but there's a useful sentence to be added which I'll do.

tturocy:
@rahulsavani I've incorporated changes based on the discussions over the last few days - pending only your final once-over before we merge.

rahulsavani (commented):
Looks good. Just one question/comment:

"The test suite makes use of a variety of games, which are drawn from examples in the game theory literature and textbooks, as well as examples constructed specifically to test edge cases. **From Gambit 17, these are all curated as part of Gambit's catalog**"

Is that true? I thought that we had decided that the catalog would only contain games of general interest (e.g. from textbooks and papers) and that games specifically constructed for testing edge cases would continue to live in `tests/test_games` or similar.

Approving, assuming this is confirmed or a corresponding minor edit is made.

tturocy:
> Looks good. Just one question/comment:
> 
> "The test suite makes use of a variety of games, which are drawn from examples in the game theory literature and textbooks, as well as examples constructed specifically to test edge cases. **From Gambit 17, these are all curated as part of Gambit's catalog**"
> 
> Is that true? I thought that we had decided that the catalog would only contain games of general interest (e.g. from textbooks and papers) and that games specifically constructed for testing edge cases would continue to live in `tests/test_games` or similar.
> 
> Approving, assuming this is confirmed or a corresponding minor edit is made.

It's a good question.

I had been influenced in drafting this by also working on merging the new features for subgames.  There, for example, we now have an example where we need an additional check in the edge case of an absent-minded information set with a single action.  On the one hand that's not going to appear in any paper, but on the other it is a useful test case that also does serve as an example of an edge case.  So that feels more catalog-y.

Are there going to be examples where we want to have test games that we load from game files but are constructed *entirely* for reasons of testing some code path in Gambit?  Entirely willing to believe the answer is "yes" to this.  I've scanned the current list of test games and part of the problem is we don't really know systematically why the games that are there, are there. :)

A good practice would be that if we have a test game, we should have a reason in the game's extended description of what it is good for.  At which point, thinking about these test games as being "circumstances that can arise in game representations that are unlikely to be covered by standard examples of game-theoretic interest but which would be useful for testing libraries which represent games" - might make them "something that ought to be in the catalog"

rahulsavani:
> > Looks good. Just one question/comment:
> > "The test suite makes use of a variety of games, which are drawn from examples in the game theory literature and textbooks, as well as examples constructed specifically to test edge cases. **From Gambit 17, these are all curated as part of Gambit's catalog**"
> > Is that true? I thought that we had decided that the catalog would only contain games of general interest (e.g. from textbooks and papers) and that games specifically constructed for testing edge cases would continue to live in `tests/test_games` or similar.
> > Approving, assuming this is confirmed or a corresponding minor edit is made.
> 
> It's a good question.
> 
> I had been influenced in drafting this by also working on merging the new features for subgames. There, for example, we now have an example where we need an additional check in the edge case of an absent-minded information set with a single action. On the one hand that's not going to appear in any paper, but on the other it is a useful test case that also does serve as an example of an edge case. So that feels more catalog-y.
> 
> Are there going to be examples where we want to have test games that we load from game files but are constructed _entirely_ for reasons of testing some code path in Gambit? Entirely willing to believe the answer is "yes" to this. I've scanned the current list of test games and part of the problem is we don't really know systematically why the games that are there, are there. :)
> 
> A good practice would be that if we have a test game, we should have a reason in the game's extended description of what it is good for. At which point, thinking about these test games as being "circumstances that can arise in game representations that are unlikely to be covered by standard examples of game-theoretic interest but which would be useful for testing libraries which represent games" - might make them "something that ought to be in the catalog"

I don't know but I could well imagine the answer is "yes", like you. On reflection, I think we can just commit to having them in the catalog -- we can always have a suitably named (sub)directory or two for these games, and it is cleaner. No such game will be totally uninteresting for the catalog, so why not?

rahulsavani (approved):

tturocy:
We can close this off by proposing the following principle:  If a game is interesting enough as a test case to be an actual game file, then it should be in the catalog - even if it's a part of the catalog which is a "rogue's gallery of edge cases".

There may be other games which we find useful in the test suite, but those we might just construct directly via the API rather than having them be saved, named games.

rahulsavani:
> We can close this off by proposing the following principle: If a game is interesting enough as a test case to be an actual game file, then it should be in the catalog - even if it's a part of the catalog which is a "rogue's gallery of edge cases".
> 
> There may be other games which we find useful in the test suite, but those we might just construct directly via the API rather than having them be saved, named games.

Agreed.

PR #913, opened 2026-06-03
Enh/hierarchical catalog

### LINKED ISSUE #912: [Catalog]: Way to view catalog via its hierarchical structure

Other

As we work on issue #901, it would be great to have a way to collapse the catalog according to the hierarchy.

Ultimately we would also like to search it this way, but actually text search in the page or the doc search function is OK for that.

For now, this would e,g. be about seeing and choosing just particlar levels in the hierarchy, e.g.

One could see just: books/geb/mor/ijgt/misc as a top level, and then within books one would see watson2013/vonstengel2022 etc.

rahulsavani:
organised via the following top level directories:

- books
- journals
- conf (for conferences)

Closes #912


rahulsavani:
@edwardchalstrey1 

Locally this did not work in the following sense:

I still had all the old entries. I think this may be because catalog/img is not cleaned out or something like that? In any case, please take a look.

@tturocy 

Are you OK with the top-level directories?

rahulsavani:
@edwardchalstrey1 @tturocy 

Separately I forgot we use these in the tests. Will fix that aspect later today.

tturocy:
> @edwardchalstrey1
> 
> Locally this did not work in the following sense:
> 
> I still had all the old entries. I think this may be because catalog/img is not cleaned out or something like that? In any case, please take a look.
> 
> @tturocy
> 
> Are you OK with the top-level directories?

Generally.  Was going to ask about your handling of Reiley et al 2008 - that's a journal article but it's in "other" as the top level.  It's a journal we're not likely to have a lot of catalog entries from, so should that be perhaps `journal/other` rather than `misc`?  `misc` seems like a good name for unpublished games that are useful but don't have a clear provenance.

tturocy (changes requested):
See comment in conversation

rahulsavani:
> > @edwardchalstrey1
> > Locally this did not work in the following sense:
> > I still had all the old entries. I think this may be because catalog/img is not cleaned out or something like that? In any case, please take a look.
> > @tturocy
> > Are you OK with the top-level directories?
> 
> Generally. Was going to ask about your handling of Reiley et al 2008 - that's a journal article but it's in "other" as the top level. It's a journal we're not likely to have a lot of catalog entries from, so should that be perhaps `journal/other` rather than `misc`? `misc` seems like a good name for unpublished games that are useful but don't have a clear provenance.

@tturocy 

Totally agree for Reiley et 2008. I am very happy with you suggestion

What about this one:

Shapley, L.S. (1974). A note on the Lemke-Howson algorithm. In: Balinski, M.L. (eds) Pivoting and Extension. Mathematical Programming Studies, vol 1. Springer, Berlin, Heidelberg

It's a (hard to find) chapter in a series that didn't continue. It's an edge case for sure.

journals/other
misc/

something else?

tturocy:
`journals/other` seems good.  Although "journal" is slightly a misnomer, but we're unlikely to have lots of games that originate in series rather than more standard journals.

edwardchalstrey1:
> Locally this did not work in the following sense:
> 
> I still had all the old entries. I think this may be because catalog/img is not cleaned out or something like that? In any case, please take a look.

@rahulsavani See the [note on step 3](https://gambitproject.readthedocs.io/en/latest/developer.catalog.html)

> You can use the `--regenerate-images` flag when building the docs locally for a second time to force any changes to be picked up.

edwardchalstrey1:
@rahulsavani I didn't make the changes directly on this branch in case you are in the midst of making changes but #915 can be merged into this branch to update the tests and docs with the new hierarchy

rahulsavani:
> @rahulsavani I didn't make the changes directly on this branch in case you are in the midst of making changes but #915 can be merged into this branch to update the tests and docs with the new hierarchy

Thanks, I sadly missed before I had done most of the work myself, but there we go. We are close now! Once this is merged it will allow me to get others contributing games effectively hopefully in the coming days, while you are still around.

edwardchalstrey1:
@rahulsavani I've updated the catalog page so there is now a collapsible structure that reflects the hierarchy you added - there's a new config called `catalog_hierarchy.yaml` which can be used to give a name to Books, Conferences, Journals and articles within journals - the names are needed so the hierarchy dropdowns are labelled in a way that makes sense. Can you check the names make sense? Page: https://gambitproject--913.org.readthedocs.build/en/913/catalog.html#catalog

rahulsavani:
> @rahulsavani I've updated the catalog page so there is now a collapsible structure that reflects the hierarchy you added - there's a new config called `catalog_hierarchy.yaml` which can be used to give a name to Books, Conferences, Journals and articles within journals - the names are needed so the hierarchy dropdowns are labelled in a way that makes sense. Can you check the names make sense? Page: https://gambitproject--913.org.readthedocs.build/en/913/catalog.html#catalog

These are perfect, at least for now. Let's get this merged in, and I will do my best to get some more games in for 16.7.

edwardchalstrey1 (approved):

edwardchalstrey1:
> > @rahulsavani I've updated the catalog page so there is now a collapsible structure that reflects the hierarchy you added - there's a new config called `catalog_hierarchy.yaml` which can be used to give a name to Books, Conferences, Journals and articles within journals - the names are needed so the hierarchy dropdowns are labelled in a way that makes sense. Can you check the names make sense? Page: https://gambitproject--913.org.readthedocs.build/en/913/catalog.html#catalog
> 
> These are perfect, at least for now. Let's get this merged in, and I will do my best to get some more games in for 16.7.

Needs an approval from @tturocy

tturocy (approved):

PR #917, opened 2026-06-03
Load OpenSpiel games via Gambit Catalog

edwardchalstrey1:
### Issues closed by this PR

Described in https://github.com/gambitproject/gambit/issues/760 but this only does OpenSpiel and only games that currently work

- This PR is an alternative to #878 that allows direct loading of OpenSpiel games compatible with (can be exported to) Gambit
- Allows the user to do `gbt.catalog.load_openspiel("tiny_hanabi")`
- Updates the documentation including OpenSpiel tutorial to load games via the catalog
- Uses mocking to test the OpenSpiel loading - users are not expected to have this installed
- Crucially, this code shouldn't need to be changed when more OpenSpiel games become exportable to NFG or EFG

### How to review this PR

- Review `load_openspiel and testing strategy
- Review explanations in catalog doc page and OpenSpiel tutorial

### DIFF
--- a/src/pygambit/catalog.py
+++ b/src/pygambit/catalog.py
@@ -49,7 +49,7 @@ def _suppress_c_stderr():
 
 def load_openspiel(game_name: str, params: dict | None = None) -> gbt.Game:
     """
-    Load a game from the OpenSpiel library into Gambit.
+    Load a game from the OpenSpiel library.
 
     Parameters
     ----------
--- a/src/pygambit/catalog.py
+++ b/src/pygambit/catalog.py
@@ -35,7 +35,7 @@ def load_openspiel(game_name: str, params: dict | None = None) -> gbt.Game:
         (e.g. ``{"players": 2, "coins": 3, "fields": 2}`` for ``"blotto"``).
         See the `OpenSpiel game list
         <https://openspiel.readthedocs.io/en/latest/games.html>`_ for
-        available parameters per game. Defaults to an empty dict.
+        available parameters per game.
 
     Returns
     -------
--- a/src/pygambit/catalog.py
+++ b/src/pygambit/catalog.py
@@ -1,4 +1,6 @@
+import contextlib
 import io
+import os
 from importlib.resources import as_file, files
 from pathlib import Path
 from typing import Any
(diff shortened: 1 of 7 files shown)


src/pygambit/catalog.py
```diff
@@ -19,6 +22,76 @@
...
+    finally:
+        os.dup2(saved_fd, 2)
+        os.close(saved_fd)
+
+
+def _load_from_openspiel(game_name: str) -> gbt.Game:
+    """
+    Load a game from the OpenSpiel library by name.
+
+    Tries NFG export first; falls back to EFG export via the
```
tturocy: I would go with exporting an efg first if it is available and only falling back to nfg, as efg is more expressive - @rahulsavani ?

src/pygambit/catalog.py
```diff
@@ -41,6 +114,9 @@ def load(slug: str) -> gbt.Game:
     """
     slug = str(Path(slug)).replace("\\", "/")
 
+    if slug.startswith(_OPENSPIEL_PREFIX):
```
tturocy: I wouldn't hook this into the main catalog calling point.  I see this more as a convenience function that we would expose directly - so just `load_openspiel` as a visible function.  This also will work better with passing arguments that control the construction of the game (see other comment)

src/pygambit/catalog.py
```diff
@@ -19,6 +22,76 @@
...
+        return
+    try:
+        yield
+    finally:
+        os.dup2(saved_fd, 2)
+        os.close(saved_fd)
+
+
+def _load_from_openspiel(game_name: str) -> gbt.Game:
+    """
```
tturocy, line 25: Looks like openspiel accepts a second argument `parameters` as a dict a a valid signature - we should mirror that so this function becomes just a simple convenience wrapper that materialises the game object in Gambit.

tturocy (changes requested):

edwardchalstrey1:
Thanks for your comments @tturocy was an easy refactor have a separate `load_openspiel` function to call  - you can now pass the params in there

src/pygambit/catalog.py
```diff
@@ -68,28 +86,27 @@ def _load_from_openspiel(game_name: str) -> gbt.Game:
...
+    # Try EFG first (works for extensive-form games).
+    # Suppress C-level stderr so any C++ error messages for games that don't
+    # support EFG export are not shown when we fall through to the NFG path.
     try:
         with _suppress_c_stderr():
-            nfg_str = pyspiel.game_to_nfg_string(game)
-        return gbt.read_nfg(io.StringIO(nfg_str))
+            efg_str = export_gambit(game)
+        return gbt.read_efg(io.StringIO(efg_str))
     except Exception:
```
tturocy: Just swallowing all exceptions is too opaque - at the moment if something goes wrong we're basically just giving a "it didn't work" result.

In particular - what happens if:
1.  The user passes a game name that isn't registered?
2.  The user passes arguments to a game that aren't supported, or misses arguments that are required (if there are any)?

We should catch and handle exceptions about unsupported export types, but not other potential errors.

tturocy (changes requested):

src/pygambit/catalog.py
edwardchalstrey1: Ok yes, I guess we just want to catch the exception raised by OpenSpiel in both these cases, but also return the correct one i.e. not the EFG generation failure if the game is normal form anyway but failed for some other reason afterwards - concatenating both errors in that case wouldn't be useful.

I should check if there's a way to distinguish normal vs extensive games loaded in OpenSpiel to begin with as that would mean we can avoid trying both.

edwardchalstrey1: @tturocy the latest commit should address the above:

> 1. The user passes a game name that isn't registered?
> 2. The user passes arguments to a game that aren't supported, or misses arguments that are required (if there are any)?

This will now just result in whatever errors `pyspiel.load_game()` throws.

I also updated the logic to use a conditional on `pyspiel.GameType.Dynamics.SEQUENTIAL` (extensive) or `pyspiel.GameType.Dynamics.SIMULTANEOUS` (normal), rather than trying to create EFG then NFG it just does one or the other.

Does this version of `load_openspiel` capture all the exceptions you were thinking?

PR #928, opened 2026-06-05
Add Wichardt 2008 (GEB) to the catalogue

d-kad:
Contributes to #394 

**Description of the changes**
This PR adds `wichardt2008.efg` to the catalogue, updates tests and the corresponding files


edwardchalstrey1:
Looks good but the game seems to not have been created on the catalog docs page - I will investigate

edwardchalstrey1:
> Looks good but the game seems to not have been created on the catalog docs page - I will investigate

Ah right @d-kad this is intentional as the EFG lacks a description - if you add one it will get picked up by the catalog. @rahulsavani do you think I should throw a warning or error for any games we try to add which lack a description, could be similar to "Make sure the game includes a description, with any citations referencing the [bibliography](https://gambitproject.readthedocs.io/en/latest/biblio.html#bibliography) using the :cite:p:`key` format, e.g. :cite:p:`Rei2008`" which is what it says in step 1 of the docs: https://gambitproject.readthedocs.io/en/latest/developer.catalog.html

d-kad:
thanks for investigating this @edwardchalstrey1! 
I will add a description now

edwardchalstrey1 (approved):

tturocy:
@edwardchalstrey1 You didn't ask me, but there's also a touching point with the discussion in #907.  

Definitely yes to having an extended description as being required for the catalog.  Basically, if a game is interesting enough to be in the catalog, it should be possible to write at least *something* about why it appears in the original source, even if it's just paraphrasing the exposition in the source.

This would also apply to "test case" games which we include in the catalog (the thrust of the discussion in #907) - in contrast to ad-hoc games we might build manually with the API in the test suite for super-specific purposes.

PR #930, opened 2026-06-05
Interface to Gamut

### LINKED ISSUE #760: [Catalog] 3rd party game libraries inclusion

### LINKED ISSUE #910: [Catalog]: Add convenience Python interface for GAMUT
We would like to offer a convenience interface to use GAMUT (http://gamut.stanford.edu) to generate games.

We envision that this would take the form of a Python module that assumes that Java and the GAMUT `.jar` file are already available and installed, and we call as an external process, similar to the model established for `PHCpack` and `lrslib`.  Functions for each of the classes of games can be generated based on the GAMUT documentation - GAMUT game families in general have many interesting parameter arguments which can be expressed as function arguments and documented in docstrings.

GAMUT generates serialised game representations, so this interface would automatically deserialise the result and return a materialised Gambit game representation.

edwardchalstrey1:
### Issues closed by this PR

- Closes #910 
- Closes #760 (SageMath to be handled later)

This PR adds:
- The ability to generate games with Gamut via the `pygambit.catalog` module, with a new function called `generate_gamut` (as oppose to `load_gamut` which would make it more similar to the OpenSpiel one, but I thought this made sense)
    - The ``params`` argument maps directly to GAMUT command-line flags. Boolean flags such as
``-normalize`` and ``-int_payoffs`` are passed as ``True``; list values for ``-actions``
expand to space-separated tokens.
- Documentation explaining how to install Gamut (and Java) and use via PyGambit - see the bottom of the [catalog page](https://gambitproject--930.org.readthedocs.build/en/930/catalog.html)
- [A tutorial notebook demonstrating using Gamut's functionality via PyGambit](https://gambitproject--930.org.readthedocs.build/en/930/tutorials/interoperability_tutorials/gamut.html), covering some of the key features of Gamut as per the documentation PDF downloaded from [here](http://gamut.stanford.edu/).
    - Some metadata has been added to prevent docs builds from attempting to run this nb, since it requires Java/Gamut, so instead the outputs are saved.

Example usage:

```python
pygambit.catalog.generate_gamut(
    "RandomGame", params={"players": 2, "actions": 3}, gamut_jar="/path/to/gamut.jar"
)
pygambit.catalog.generate_gamut(
    "CovariantGame",
    params={"players": 2, "actions": [3, 3]},
    gamut_jar="/path/to/gamut.jar",
)
pygambit.catalog.generate_gamut(
    "RandomGame",
    params={
        "players": 3,
        "actions": 4,
        "normalize": True,
        "min_payoff": 0,
        "max_payoff": 100,
    },
    gamut_jar="/path/to/gamut.jar",
)
```

Todo:
- [x] Review and edit the tutorial, ensuring it makes sense and covers what a user might find helpful
- [x] Update tutorial or docs to cover all the available GAMUT classes

### How to review this PR

@tturocy @rahulsavani I've taken an approach to this that mirrors that of OpenSpiel loading since #910 is listed as a catalog issue, however re-reading I notice Ted said "and we call as an external process, similar to the model established for `PHCpack` and `lrslib`" - so perhaps that approach makes more sense than what I've done here. Since Gamut involves Game generation rather than just loading pre-made games as in OpenSpiel (though I suppose you can pass in params there too)... my thought is perhaps Gamut generation should live elsewhere from the Catalog module? Again this brings up the issue of the Catalog being currently a PyGambit-only feature which Ted has raised. What are your thoughts?


doc/tutorials/interoperability_tutorials/gamut.ipynb
```diff
@@ -0,0 +1,534 @@
+{
```
edwardchalstrey1, line 1: <p>Is this accurate for the definition of the one MSP above?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/gambitproject/gambit/pull/930/discussion/#comment-3412485899'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='8'/>

tturocy:
This looks quite good to me already out of the box.

I still have a feeling that the verb differentiation between OpenSpiel and GAMUT here isn't meaningful - I could just as easily justify "generate" for OpenSpiel, or "load" for both.  In both cases the external tools are doing a generation step, even if we are using .efg/.nfg as the communication format those files are being created on demand by the library we're calling.

edwardchalstrey1:
> This looks quite good to me already out of the box.
> 
> I still have a feeling that the verb differentiation between OpenSpiel and GAMUT here isn't meaningful - I could just as easily justify "generate" for OpenSpiel, or "load" for both. In both cases the external tools are doing a generation step, even if we are using .efg/.nfg as the communication format those files are being created on demand by the library we're calling.

Shall we stick with "load" for both to be consistent with the regular catalog load function?

tturocy:
I can see it both ways.

A reason I might lean to `generate` is the distinction that those methods may do work to realise the game beyond simply reading an already-serialised format.  It can be useful to have different verbs or syntax in the API to hint at the underlying complexity or process.  Compare for example the principle with Python `@property` that it's not good style to hide something that's computationally non-trivial in a property, just because then the API is communicating information that calling code should be aware of.

So `load` would mean "read an existing serialisation" while `generate` would mean "execute some code that constructs the representation"

edwardchalstrey1:
Thanks @tturocy `load_openspiel` has now been renamed to `generate_openspiel` as per your comment above

edwardchalstrey1:
Hi @wyz2368 I was just discussing with @rahulsavani - it would be helpful to have your review on this pull request. If you could check out the branch and read and run the tutorial I added locally (noting you will need to download Gamut as it should explain) and let me know if anything doesn't work, and also if the tutorial makes sense or should be re-written at all.

wyz2368:
@edwardchalstrey1 I tested all the examples in the notebook and everything works. Thank you.

edwardchalstrey1:
Thanks @wyz2368 !

doc/tutorials/interoperability_tutorials/gamut.ipynb
```diff
@@ -0,0 +1,809 @@
...
```
rahulsavani, line 1: <p>Unfortunately, this is not a standard version of Battle of the Sexes (not your fault, Gamut's). The standard version would have by symmetric, but only by swapping players AND actions. However this game in Gamut is unusually actually "symmetric" in the standard game-theoretic sense with payoffs (3,3) for (Opera, Opera) and likewise for (Football, Football). A standard (non-symmetric) version, e.g. , a la Wiki:</p><p> </p><p><a href="https://en.wikipedia.org/wiki/Battle_of_the_sexes_(game_theory)" target="_blank">https://en.wikipedia.org/wiki/Battle_of_the_sexes_(game_theory)</a></p><p> </p><p>has the man preferring (Football, Football) to (Opera, Opera) and vice versa for the woman, e.g. with payoffs (3, 2) and (2, 3). We should avoid confusing people.</p><p> </p><p><span class="mention" data-value="tturocy" data-denotation-char="@">﻿<span contenteditable="false">@tturocy</span>﻿</span> Do you agree?</p><p> </p><p>If so, a simple fix would be to swap out BoS for Chicken, which is symmetric in Gamut and symmetric as standard too.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/gambitproject/gambit/pull/930/discussion/#comment-3460787245'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='4'/>

rahulsavani (changes requested):
Looks good apart from the fact Gamut (and thereby we) have used a non-standard version of Battle of the Sexes. We should probably swap it out for, e.g. Chicken. Let's see if @tturocy agrees.

tturocy:
Yes, I would agree with using Chicken instead, doesn't require substantial changes to the exposition of the example but users familiar with game theory won't be surprised by the game's payoffs.

rahulsavani (approved):
Looks good to me.

PR #942, opened 2026-06-12
Remove integer indexing from game collections in `pygambit`

d-kad:
### Description 

Game object collections in `pygambit` are now indexed by string label only.
- Integer indexing is removed from `Game.players`, `Game.outcomes`, `Game.strategies`, `Game.infosets`,
  `Game.actions`, `Player.strategies`, `Player.infosets`, `Player.actions`, `Infoset.actions`, `Infoset.members`,
   and `Node.children`. Indexing by an integer raises `TypeError`. 
   Positional access remains available by iterating a collection (e.g. `pl1, pl2 = game.players`).
   Indexing a `Game` by a contingency (e.g. `game[0, 1]`) is unchanged.
- Label lookup is by exact match.
  A label matching no object uniformly raises `KeyError` (`Node.children` previously raised `ValueError`).
- Lookup is centralised in a new `_resolve_by_label` helper; all the collection `__getitem__` methods use it.
   The `label` parameter is type-hinted `str`, so Cython rejects non-string indices at the call boundary.
- The test suite (including the tutorial notebooks) is fully migrated to label-based access (~460 affected tests),
   with new tests added pinning the integer-rejection across all collections 
   (`COLLECTION_GETTERS` in `tests/test_game.py`).


d-kad:
**Design question — should label lookup strip whitespace, or match exactly?** 

Before this PR, the collection `__getitem__`s stripped the *lookup string* before comparing:

```python
matches = [x for x in self if x.label == index.strip()]
```

This appears at 10 sites (`game.pxi` ×5, `player.pxi` ×3, `infoset.pxi` ×2),  
so e.g. `g.players[" Alice "]` would find the player labelled `"Alice"`. 
The most recent new `Node.children` semantics in #588, matches labels verbatim. 

In this PR I matched labels **exactly**. ~~ *(superseded — see update)*

**UPD:** @rahulsavani and I agreed label lookup should keep stripping surrounding
whitespace from the query (stored labels are matched as-is).

src/pygambit/game.pxi
```diff
@@ -300,19 +300,30 @@ class GameOutcomes:
...
-            matches = [x for x in self if x.label == index.strip()]
-            if not matches:
-                raise KeyError(f"Game has no outcome with label '{index}'")
-            if len(matches) > 1:
-                raise ValueError(f"Game has multiple outcomes with label '{index}'")
-            return matches[0]
-        if isinstance(index, int):
-            return Outcome.wrap(self.game.deref().GetOutcome(index + 1))
-        raise TypeError(f"Outcome index must be int or str, not {index.__class__.__name__}")
+    def __getitem__(self, label) -> Outcome:
```
tturocy: For this and all other `__getitem__`: If we use `label: str` with the explicit type hint, then Cython will automatically do the type checking and generate an exception.  So we should have type hints on all of these.

src/pygambit/gambit.pyx
```diff
@@ -74,6 +73,44 @@ def _to_number(value: typing.Any) -> c_Number:
     return c_Number(value.encode("ascii"))
 
 
+@cython.cfunc
```
tturocy, line 76: I believe we can add `label: str` as a type hint here if we also do it on all of the `__getitem__` calls (see other comment)

tturocy (changes requested):
A few places where we should be using Cython type hints.

tturocy (approved):

PR #947, opened 2026-06-17
[GUI] Create outcome editor panel for extensive forms

tturocy:
This changes the UX for creating and editing outcomes in the extensive form.

Previously, we attempted to edit payoffs in the extensive form by popping up a text control more or less in the "right" place on the extensive form canvas, and allowing edits via e.g. tab traversal.  This had a number of poor features:

* It was quite brittle and doesn't work very well with having to check for and validate "unsaved changes" in an edit.
* It did not expose outcome labels - which we now consider to be important (in fact essential)
* Even when editing only one payoff, it still makes sense to be thinking about the outcome as being associated with a vector of payoffs.

This replaces the editing mechanism with a popup window which displays the outcome label and all payoffs.  Edits are accepted by `Enter` or clicking elsewhere in the parent window, and are dismissed via `Esc`, so changes to the outcome are atomic at the edit level (rather than the individual payoffs level as before).


tturocy:
All, this is also still WIP but testing and suggestions are very welcomed.

edwardchalstrey1:
Working for me, though I wonder if there's anything you can do to make it clearer that double click lets you edit the outcomes. Also, it then made me think "why can't I double click to edit action labels"

Screenshot 2026-06-18 at 11 50 42

d-kad:
looks good to me -- one small thing to flag, likely outside this PR's scope: payoff input doesn't normalize negative zero
Screenshot 2026-06-18 at 12 24 47

Issue #862, opened 2026-04-29
[ENH]: Support unicode for labels


tturocy:
Gambit almost (but not quite) predates Unicode, but certainly the main codebase predates good support for Unicode across platforms.

We should ensure that non-ASCII characters can be used in object labels throughout.  The recommendation would be that we assume UTF-8, and insist only on byte-wise equality (rather than normalised equality) as far as uniqueness of labels (within the scopes in which labels are now required to be unique).

A known limitation is that at the moment encoding/decoding of labels in `pygambit` uses ASCII, so manipulating labels with non-ASCII characters definitely does not work in Python at present.

The graphical interface is somewhat more supportive of non-ASCII characters, but a thorough check/audit should be done to ensure labels are handled consistently everywhere.

tturocy:
We note that in #944 and #945 we are making some changes which, in version 16.7, will enforce that labels must be ASCII-only for now.  In practice anything non-ASCII does not work reliably at the moment - sometimes such labels will work, sometimes not.

In implementing this issue we will ensure that the validation and normalisation being done in #944 and #945 is updated to support Unicode.  The idea is that by having a small number of sites where validation and normalisation are done, adding Unicode support across Gambit should then be easy.

Issue #945, opened 2026-06-16
[ENH]: Improve UX in graphical interface for valid labels


tturocy:
We are now defining some restrictions on valid labels, see #944.

As part of this we should add some UX to help do some simple normalisations in the graphical interface:
* Stripping leading and trailing whitespace
* Normalising multiple spaces into a single space
* Accepting for the moment only ASCII characters.

On the last point we note that #862 will introduce proper Unicode support for 17.0.  We are making a conscious decision temporarily to suspend Unicode input - because it does not work properly at the moment, and it is our judgment that we should set up the proper structure for defining legal labels first and that will make it easier to have valid Unicode labels for 17.0.

PR #957, opened 2026-06-22
Define and enforce a valid-label criterion for game objects (#944)

### LINKED ISSUE #944: [ENH]: Define and enforce criteria for legal labels in game representation
Now that we insist on labels for many objects within a game, we should define what a legal label is.

For version 16.7, we propose that a legal label satisfies the following:
* Consists only of printable ASCII characters or spaces
* The first and last character of the label must be printable
* Other whitespace or non-printable characters are *not* acceptable
* Two or more consecutive spaces are not permitted

This should be added as a check and enforced at the C++ level, to ensure uniformity.  This will have implications for the checks in pygambit (see discussion in PR #942).

We also want to centralise this checking because in 17.0 we will allow Unicode labels more generally (see #862).  At the moment Unicode labels sometimes work but largely by accident if they do.  Therefore our approach will be first to ensure validity of labels and then this will make it easier to expand that definition (as opposed to our current approach which is inconsistent across different sites).

d-kad:
### Issues closed by this PR
- Closes #944

This PR defines and enforces a single criterion for valid game-object labels, centrally at the C++ level.

A label is valid if it consists only of printable ASCII characters and spaces, does not begin or end with a space, and does not contain two consecutive spaces. The empty label remains valid. The definition is enforced in one place (`IsValidLabel`/`CheckLabel` in `game.h`) so that widening it to permit Unicode in 17.0 (#862) is a localised change.

- `CheckLabel` is called from all six `SetLabel` methods (outcome, action, infoset, strategy, player, node) and from the `GameStrategyRep` constructor.
- In `pygambit`, the six `SetLabel` declarations and `NewStrategy` now map the C++ exception to `ValueError` (`except +ValueError`); an invalid label raises `ValueError`.
- `add_strategy` now encodes the label as ASCII, consistent with the other accessors; a non-ASCII label fails at the boundary with `UnicodeEncodeError`.

Tests add per-type valid/invalid/non-ASCII coverage (constants in `tests/games.py`).

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -39,6 +39,8 @@
 - Payoff editing in extensive games in the graphical interface is now done via a context popup window
   rather than text controls drawn (not always well!) over the game tree display.  (#947)
 - In `pygambit`, indexing game object collections by integer position has been removed.  (#942)
+- Validity of game object labels is enforced (printable ASCII and spaces only, no leading/trailing or
+  double spaces); invalid labels raise `ValueError` in `pygambit`. (#944)
 
 ### Removed
 - Built-in plotting of logit QRE for strategic games has been removed in the GUI (#809)
--- a/src/games/game.h
+++ b/src/games/game.h
@@ -112,6 +112,58 @@ public:
 };
 
 //=======================================================================
+//                       Validation of labels
+//=======================================================================
+
+/// @brief Returns whether p_label is a valid label for a game object.
+///
+/// A valid label either is the empty string (denoting the absence of a label),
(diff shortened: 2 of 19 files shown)


tturocy:
@d-kad This looks good for a start.  I did check and there are some paths in the GUI which aren't catching exceptions, so if we merge this without adding at least some crude defensive handlers we're creating a bad regression (it completely crashes the GUI!) so I need to look at those first before we can accept (unfortunately we can't treat them entirely as separate issues as #945 wishes were true!)

PR #962, opened 2026-06-25
Make `Node.infoset`, `Node.outcome`, `Node.player` lazy, node-anchored views (#946)

### LINKED ISSUE #946: [API]: Lazy evaluation of all relations between game elements
As part of developing our "algebra of games" idea, we want to move towards "lazy evaluation" of relationships among the constituent parts of games.  See content migrated from discussion below for the rationale of why this makes the API more consistent.

For the moment we will change this in `pygambit` only (with C++ intended for 17.0 alongside more substantial implementation changes).

We will need to change (at least) the implementation of the following:
* `Node.outcome`
* `Node.infoset`
* `Node.player`
* `Node.parent`

It is not a coincidence that these are all node-related properties.  We are intending that the node (history) is going to be the principal way of manipulating extensive games in 17.0, with objects referenced relative to a node (or set of nodes).  

The implementation path for this in `pygambit` would be creating a different type of object which instead of keeping, for example, an infoset, will keep a node, and then access the node's information set on demand when needed.

### Discussed in https://github.com/gambitproject/gambit/discussions/863

<div type='discussions-op-text'>

<sup>Originally posted by **tturocy** April 30, 2026</sup>
This is something which is envisaged as part of the algebra-of-games API planned for 17.0.

Presently, relations like `Node.outcome` or `Node.infoset` are evaluated "immediately" and resolve to an object which is realised physically in memory (ultimately it's a shared pointer to an allocated structure).  We have developed a sense that we want to move away from this, and to treat these as a form of selector.

There is another justification for this, which is that the existing API is inconsistent in this regard.  Consider the sequence below:

```python
In [1]: import pygambit as gbt
g = gbt.
In [2]: g = gbt.Game.new_tree(players=["1", "2"])

In [3]: first_gen = g.root.children

In [4]: len(first_gen)
Out[4]: 0

In [5]: g.append_move(g.root, "1", actions=["U", "D"])

In [6]: len(first_gen)
Out[6]: 2
```

As implemented, `Node.children` actually returns an object which is evaluated, in effect, lazily on iteration.  So if the set of children of the node change, the output of iterating the collection changes.

However, predicates that resolve to individual objects rather than collections are evaluated immediately.  For example,

```python
In [1]: import pygambit as gbt

In [2]: g = gbt.Game.new_tree(players=["1", "2"])

In [3]: infoset = g.root.infoset

In [4]: infoset

In [5]: g.append_move(g.root, "1", actions=["U", "D"])

In [6]: infoset == g.root.infoset
Out[6]: False
```

This is (in my view) an argument in favour of revising the semantics even for 16.7 for these properties.  I claim that *most* of the time, the intention when referring to objects via these relations is in the sense of a predicate.  `g.root.infoset` means "the information set to which the root node of the game belongs", which may change as the game mutates; `g.root.outcome` would mean "the outcome associated with the root node (if any)", which likewise might change.
</div>

d-kad:
Closes #946

Makes three node properties — `Node.infoset`, `Node.outcome`, and `Node.player` — return
lazy, node-anchored views (`NodeInfoset`, `NodeOutcome`, `NodePlayer`) that resolve on each
access rather than capturing a value at access time. A stored reference now tracks game
mutations and re-resolves to reflect the current tree, consistent with the already-lazy `Node.children`.

Each view resolves to `None` (and is falsy) when the relation is absent: a terminal node has
no infoset and no player; same for a node with no attached outcome. At a chance node,
`Node.player` resolves to the chance player (`is_chance` is `True`). 

`Node.parent` remains eager as agreed.

### Notes

- The views are hashable (by their resolved object); transitional. No public `resolve()`.
- `Infoset.__eq__`, `Outcome.__eq__`, and `Player.__eq__` now return `NotImplemented` on type
  mismatch (was plain `False`), so a real object and its view compare equal from either side.


src/pygambit/node.pxi
```diff
@@ -95,6 +95,94 @@ class NodeChildren:
...
+        obj.node = node
+        return obj
+
+    @cython.cfunc
+    def _resolve(self) -> Infoset:
+        if self.node.deref().GetInfoset() == cython.cast(c_GameInfoset, NULL):
+            return None
+        return Infoset.wrap(self.node.deref().GetInfoset())
+
+    def resolve(self) -> Infoset:
```
tturocy: I would suggest we don't want this as a public operation - the idea is to get away from this way of thinking because this is in essence treating a specific C++ structure in memory as being an information set, rather than the concept of "the set of nodes which are information-equivalent to this one from the perspective of the DM"

src/pygambit/node.pxi
```diff
@@ -95,6 +95,94 @@ class NodeChildren:
...
+        raise ValueError("Cannot create a NodeInfoset outside a Game.")
+
+    @staticmethod
+    @cython.cfunc
+    def wrap(node: c_GameNode) -> NodeInfoset:
+        obj: NodeInfoset = NodeInfoset.__new__(NodeInfoset)
+        obj.node = node
+        return obj
+
+    @cython.cfunc
```
tturocy, line 119: Having this as a private function is what we do need!

src/pygambit/node.pxi
```diff
@@ -153,15 +241,19 @@ class Node:
         return Game.wrap(self.node.deref().GetGame())
 
     @property
-    def infoset(self) -> Infoset | None:
-        """The information set to which this node belongs.
+    def infoset(self) -> NodeInfoset:
```
tturocy, line 396: I think I see why this change.  The `infoset(node)` can change - so at a given moment the infoset might be non-null but then become null if the tree is truncated.  So we always need to return the wrapper object.

tests/test_node.py
```diff
@@ -534,12 +554,13 @@ def test_insert_move_error_player_mismatch():
...
 
 def test_node_leave_infoset():
-    """Test to ensure it's possible to remove a node from an infoset"""
+    """A node-anchored infoset proxy is lazy: it re-resolves after the node leaves its infoset."""
     game = games.read_from_file("basic_extensive_game.efg")
-    p2_infoset = game.root.children["U1"].infoset
-    assert len(p2_infoset.members) == 2
-    game.leave_infoset(game.root.children["U1"])
-    assert len(p2_infoset.members) == 1
+    node = game.root.children["U1"]
```
tturocy, line 597: So this shows that very few tests were broken by this change - and that the test was written in kind of a not-good way to start with!

src/pygambit/node.pxi
```diff
@@ -95,6 +95,94 @@ class NodeChildren:
         raise TypeError(f"Index must be a str label or an Action, not {action.__class__.__name__}")
 
 
+@cython.cclass
```
tturocy, line 98: As the tests show we need to be able to `__hash__` a `NodeInfoset`, at least for now.  This might be a transitional requirement depending on how the algebra of games goes.

tturocy (changes requested):

tturocy (changes requested):
This looks positive, so let's proceed with `.outcome` and `.player` as discussed and combine them into one atomic change.

tturocy (approved):

PR #966, opened 2026-07-01
Improved UX for editing labels in GUI

### LINKED ISSUE #945: [ENH]: Improve UX in graphical interface for valid labels

tturocy:
This adds some improved UX in the GUI for inputting labels:
* Leading whitespace and multiple whitespace are normalised on the fly while editing
* Trailing whitespace is normalised transparently on commit
* For now, non-ASCII input is filtered on the fly; this has been written to make it easy to switch to Unicode for 17.0 as planned but aligns with the current rules.

Closes #945.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -475,6 +475,8 @@ gambit_SOURCES = \
 	src/gui/editlabel.h \
 	src/gui/edittext.cc \
 	src/gui/edittext.h \
+	src/gui/labelcell.cc \
+	src/gui/labelcell.h \
 	src/gui/dlabout.cc \
 	src/gui/dlabout.h \
 	src/gui/dleditmove.cc \
--- a/Makefile.am
+++ b/Makefile.am
@@ -471,6 +471,8 @@ gambit_SOURCES = \
 	src/gui/analysis.h \
 	src/gui/app.cc \
 	src/gui/app.h \
+	src/gui/editlabel.cc \
+	src/gui/editlabel.h \
 	src/gui/edittext.cc \
 	src/gui/edittext.h \
 	src/gui/dlabout.cc \
--- a/src/gui/dleditmove.cc
+++ b/src/gui/dleditmove.cc
@@ -46,6 +46,8 @@ public:
 
   wxString GetActionLabel(int p_act) const;
   Array<Number> GetActionProbs() const;
+
+  void FocusActionLabel(int p_act) { m_actionLabels.at(p_act - 1)->SetFocus(); }
 };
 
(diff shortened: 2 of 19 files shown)


tturocy:
@rahulsavani @d-kad If you can try out the various places we set/edit labels that would be helpful.

d-kad:
Tested on macOS

Action labels work fine (you can see that action label turned "Éxït" into "xt" and removed trailing spaces from the second label automatically)

Node labels through on invalid input
Screenshot 2026-07-01 at 15 48 49
Screenshot 2026-07-01 at 15 48 54

tturocy:
I wasn't able to reproduce this with the node labels; I did change that editor and on mine I'm not able to input invalid labels.  Did you build yourself from the branch?  If so are you sure you've got the latest version of it?

d-kad:
You're right — I was three commits behind on issue_945, everything works as expected now. 

Screenshot 2026-07-01 at 20 29 42

Sorry for raising a false alarm!

d-kad:
## Label editing test pass

### Move dialog (infoset + action labels)
- Typed `  Go   Left  ` in an action label → becames `Go Left`
- Typed and pasted `é` / emoji → filtered silently
- Two actions in same infoset given same label → error, dialog stays open
- Action 2 = `Left ` (trailing space) when action 1 = `Left` → still rejected
- Empty action label → blocked
- Same infoset label on both of P1's infosets → error; same label on P2's infoset → allowed
- Empty infoset label → allowed
- In the chance infoset probabilities are editable

### Node dialog
- Duplicate an existing node label → error
- Typed `Node1   ` then hit Enter (not the OK button)
- `é` filters 

### Player names (EFG and NFG)
- Rename P2 to P1's name → error
- Empty name → blocked; Esc mid-edit → reverts
- `  P   1  ` → normalizes

### Strategy names (NFG)
- Duplicate strategy label within one player → error
- Same strategy label as the *other* player's strategy → allowed
- Empty strategy label → blocked

d-kad (approved):

PR #967, opened 2026-07-01
Add shell infrastructure for Herings-Peeters (2001) solver

AndresFerCervell:

This PR implements the basic shell infrastructure for a new Nash equilibrium solver based on the differentiable homotopy method by Herings and Peeters (2001).

Changes included:
- **C++ Layer (`src/solvers/hp/`)**: Created `hp.h` and `hp.cc`, currently initializes a `StrategySupportProfile` and returns a trivial uniform mixed strategy profile.
- **Cython Layer (`src/pygambit/`)**: Added external C++ declarations to `gambit.pxd` under the `Gambit` namespace, and implemented the `_hp_strategy_solve` translation function inside `nash.pxi`.
- **Python Layer (`src/pygambit/nash.py`)**: Exposed the high-level `hp_solve` function to end-users.
- **Build System**: Updated `setup.py` and updated `Makefile.am` for general consistency.
- **Documentation (`doc/references.bib`)**: Added the canonical BibTeX entry for Herings & Peeters (2001) under the `articles_equilibria` category. Documentation will be expanded throughthe creation of the algorithm.
### How to review this PR

-Ensure the project compiles succesfully.
-Verify that running the solver from a Python console returns the expected trivial profile. 
-Check that the internal naming schemas adhere to Gambit´s established conventions for strategic form solvers.


src/solvers/hp/hp.cc
```diff
@@ -0,0 +1,38 @@
...
+// along with this program; if not, write to the Free Software
+// Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA 02111-1307, USA.
+//
+
+#include <iostream>
+#include "gambit.h"
+#include "solvers/hp/hp.h"
+
+namespace Gambit {
+std::list<MixedStrategyProfile<double>> HPStrategySolveWrapper(const Game &p_game)
```
tturocy: I think we can just call this `HPStrategySolve` (but see separate comments on naming)

Makefile.am
```diff
@@ -207,6 +207,8 @@ EXTRA_DIST = \
 	contrib/games/yamamoto.nfg \
 	contrib/games/zero.nfg \
 	src/README.rst \
+	src/solvers/hp/hp.cc \
```
tturocy: Here we should add a new group of files to parallel the other solvers.  `EXTRA_DIST` is by convention for non-source-code files; these should be in a `<something>_SOURCES` declaration.

We don't need to write the C++ driver (as discussed, as this will be obsoleted in 17.0), but still will want to group the files required for each method.

tturocy (changes requested):
The initial setup looks good.

We should consider how to organise this.  I am thinking we will set up a long-running branch for development of this (separate from `master`) and then do PRs against that branch.  That way we can keep locking in logical steps of progress without winding up with a single huge PR to resolve down the line.

tturocy:
@rahulsavani @d-kad For thought/discussion: What should we call this method?  We have not added new methods for quite some time.  There are several more we are planning to add.

In the past we have avoided referring to methods by author names.  However, all methods had a reasonably natural naming in terms of their method.  Possibly the exception would be `simpdiv`, which is indeed a simplicial subdivision method but not necessarily (?) the unique possibility for such a method.  

Herings-Peeters were indeed the first to publish a differential path-following method implemented as a true path-following method.  But as noted H-P is in practice an implementation of Harsanyi-Selten, and Logit QRE likewise pre-dated H-P (although again my formulation of it as path-following came later).

Views are welcomed - for now @AndresFerCervell carry on calling this "HP" and the like while we think about it!

tturocy (approved):

tturocy:
I've created a branch `feature/hp` in the main repository which we'll use as the base for work on this over the coming weeks.  That will help keep it insulated from other planned changes which will be underway on `master` once we release 16.7.

PR #969, opened 2026-07-02
Add 3 player max cut test for enumpoly solver

ameliekleber:
This PR adds a test for the enumpoly solver that covers the 2x2x2 max cut normal form game, ensuring it returns the two pure and one mixed equilibirum.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -102,6 +102,7 @@ markers = [
     "nash_enumpure_strategy: tests of enumpure_solve in pure strategies",
     "nash_enumpure_agent: tests of enumpure_solve in pure behaviors",
     "nash_enummixed_strategy: tests of enummixed_solve in mixed strategies",
+    "nash_enumpoly_strategy: tests of enumpoly_solve in mixed strategies",
     "nash_enumpoly_behavior: tests of enumpoly_solve in mixed behaviors",
     "nash_lcp_strategy: tests of lcp_solve in mixed strategies",
     "nash_lcp_behavior: tests of lcp_solve in mixed behaviors",
--- a/tests/test_games/2x2x2_nfg_from_local_max_cut_2_pure_1_mixed_eq.nfg
+++ b/tests/test_games/2x2x2_nfg_from_local_max_cut_2_pure_1_mixed_eq.nfg
@@ -6,6 +6,7 @@ NFG 1 R "2x2x2 game with 2 pure and 1 mixed equilibrium"
 - Pure strategies {a,b} encode if respective player is on left or right of the cut
 - The payoff to a player is the sum of their incident edges across the implied cut
 - Pure equilibrium iff local max cuts; in addition, uniform mixture is an equilibrium
+- In the mixed equilibrium all players mix strategies with equal probability (0.5, 0.5)
 - Equilibrium analysis for pure profiles:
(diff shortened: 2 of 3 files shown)


ameliekleber:
@rahulsavani I added the test we discussed on Wednesday, would appreciate your review :)

rahulsavani (changes requested):
Thanks, this looks good to me, though co-pilot is right, we should probably include a regret_tol too. This max cut game is super simple so it may never error, but it's still good practice. For the suggestion below with nau2004, we'll definitely need that anyway.

Further suggestion: 

Sorry to be demanding but I didn't realise we had no tests at all for "ENUMPOLY_STRATEGY", could you perhaps add a couple more, e.g. a 2-player coordination games would be natural (say in dimension 4 where we expect 7 eq, assuming it runs quickly enough, or dim 3 if not), for one or two further 3-player games you could try:

```
gbt.catalog.load("journals/ijgt/nau2004/sec4")
gbt.catalog.load("journals/ijgt/nau2004/sec5")
```

If sec5 takes too long for stop_after=None, feel free to set it otherwise. 

Thanks.

tturocy:
@rahulsavani Further to our in-person conversation, indeed it looks like we never got around to doing strategy versions of tests for `enumpoly` - likely due to combination of issues with coordination games and with ordering of outputs...!

ameliekleber:
I've added the 3 test you requested, the two from the catalog and the coordination game for n=3 (I think this is what you meant as this one has 7 eq and n=4 has 15?). The coordination game test was already in the behavior tests for enumpoly, so I added the use_strategic parameter for this test

rahulsavani:
> I've added the 3 test you requested, the two from the catalog and the coordination game for n=3 (I think this is what you meant as this one has 7 eq and n=4 has 15?). The coordination game test was already in the behavior tests for enumpoly, so I added the use_strategic parameter for this test

You are right I have an "off by one error" with my reference to the dimensions of the coordination game: n=3 is good.

rahulsavani:
https://github.com/all-contributors

please add @ameliekleber  for code

rahulsavani:
@tturocy 

```
    # A three-player 2x2x2 game with 3 pure, 2 incompletely mixed, and a continuum of completely mixed Nash equilibria (nau2004 sec5 catalog game)
    pytest.param(
        EquilibriumTestCase(
            factory=functools.partial(gbt.catalog.load, "journals/ijgt/nau2004/sec5"),
            solver=functools.partial(gbt.nash.enumpoly_solve, stop_after=None),
            expected=[
                    [d(1, 0), d(0, 1), d(1, 0)],
                    [d(0, 1), d(1, 0), d(1, 0)],
                    [d(0, 1), d(0, 1), d(0, 1)],
            ],
            prob_tol=TOL,
            regret_tol=TOL,
        ),
        marks=pytest.mark.nash_enumpoly_strategy,
        id="test_enumpoly_strategy_4",
```

This is more like a regression test because I am not sure that we should only get the pure here -- or is this clear to you?

rahulsavani (changes requested):
Looks good. 

One minor thing is flake8 -- please install pre-commit, which will help with this.

For a quick look at the required changes in this case, please see what co-pilot did in PR #972

rahulsavani (approved):

PR #979, opened 2026-07-08
Enforce unique, nonempty labels for players

d-kad:
### Issues closed by this PR
- Part of #861

Enforces the label rules for players, following #978 (nodes and information sets). Players are a non-empty, unique-within-game scope, enforced in `GamePlayerRep::SetLabel`:

- A player label must be non-empty, valid, and unique among the game's players; a violating label raises `ValueError`.
- The chance player is labeled `"Chance"`, set through the constructor, and its label is reserved: `SetLabel` rejects any attempt to relabel it.
- `add_player` now requires a label (no default, no auto-generation).

### DIFF
--- a/src/games/game.h
+++ b/src/games/game.h
@@ -1350,6 +1350,9 @@ inline void GamePlayerRep::SetLabel(const std::string &p_label)
     return;
   }
   CheckLabel(p_label);
+  if (GetGame()->IsTree() && p_label == GetGame()->GetChance()->GetLabel()) {
+    throw ValueException("Player label must not be the reserved chance player label");
+  }
   for (const auto &player : GetGame()->GetPlayers()) {
     if (player.get() != this && player->GetLabel() == p_label) {
       throw ValueException("Player label must be unique within the game");
--- a/src/games/game.h
+++ b/src/games/game.h
@@ -464,7 +464,7 @@ public:
     : m_game(p_game), m_number(p_id), m_label(p_label)
   {
   }
-  GamePlayerRep(GameRep *p_game, int p_id, int m_strats);
+  GamePlayerRep(GameRep *p_game, int p_id, const std::string &p_label, int p_strats);
   ~GamePlayerRep();
 
   bool IsValid() const { return m_valid; }
--- a/src/games/game.h
+++ b/src/games/game.h
@@ -460,7 +460,10 @@ public:
   using Strategies = ElementCollection<GamePlayer, GameStrategyRep>;
   using Sequences = ElementCollection<GamePlayer, GameSequenceRep>;
 
-  GamePlayerRep(GameRep *p_game, int p_id) : m_game(p_game), m_number(p_id) {}
+  GamePlayerRep(GameRep *p_game, int p_id, const std::string &p_label)
+    : m_game(p_game), m_number(p_id), m_label(p_label)
+  {
+  }
   GamePlayerRep(GameRep *p_game, int p_id, int m_strats);
   ~GamePlayerRep();
 
@@ -471,11 +474,7 @@ public:
   Game GetGame() const;
 
   const std::string &GetLabel() const { return m_label; }
-  void SetLabel(const std::string &p_label)
-  {
-    CheckLabel(p_label);
-    m_label = p_label;
-  }
+  void SetLabel(const std::string &p_label);
 
   bool IsChance() const { return (m_number == 0); }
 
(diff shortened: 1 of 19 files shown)


tturocy:
I have pushed some fixes to the GUI.  The previous changes were assuming that "Player (n+1)" would necessarily be a unique label, which it might not be.  This corrects that (and centralises the unique label generation to a single site).

In the Python tests, we don't have a test to confirm that trying to create a player with a duplicate or empty label fails, and further that doing so leaves the game unchanged.

src/pygambit/game.pxi
```diff
@@ -2053,23 +2053,40 @@ class Game:
...
+        ------
+        ValueError
+            If `label` is empty, is already the label of another player, or (in an
+            extensive game) is ``"Chance"``, the reserved label of the chance player.
         """
-        p = Player.wrap(self.game.deref().NewPlayer())
-        if str(label) != "":
-            p.label = str(label)
-        return p
+        label_str = str(label)
```
tturocy: I'm not sure what calling `str()` on what is already guaranteed to be a `str()` is intended to do?

src/pygambit/game.pxi
```diff
@@ -2053,23 +2053,40 @@ class Game:
...
+        if not label_str:
+            raise ValueError("add_player(): label must not be empty")
+        existing = [player.label for player in self.players]
```
tturocy: All these checks duplicate what is already supposed to be done in `NewPlayer`.  And in fact they mask a bug: C++ `NewPlayer` allows the creation of a player with the label "Chance".

src/pygambit/player.pxi
```diff
@@ -256,11 +256,6 @@ class Player:
 
     @label.setter
     def label(self, value: str) -> None:
-        if value == self.label:
-            return
-        if value == "" or value in (player.label for player in self.game.players):
-            warnings.warn("In a future version, players must have unique labels",
-                          FutureWarning)
         self.player.deref().SetLabel(value.encode("ascii"))
```
tturocy, line 259: Unlike in `add_player` above, here we rely on the logic already being correct in C++ and don't duplicate it in Cython.  This is better.

src/games/game.h
```diff
@@ -460,7 +460,10 @@ class GamePlayerRep : public std::enable_shared_from_this<GamePlayerRep> {
   using Strategies = ElementCollection<GamePlayer, GameStrategyRep>;
   using Sequences = ElementCollection<GamePlayer, GameSequenceRep>;
 
-  GamePlayerRep(GameRep *p_game, int p_id) : m_game(p_game), m_number(p_id) {}
+  GamePlayerRep(GameRep *p_game, int p_id, const std::string &p_label)
+    : m_game(p_game), m_number(p_id), m_label(p_label)
+  {
+  }
   GamePlayerRep(GameRep *p_game, int p_id, int m_strats);
```
tturocy: To be consistent both ctors should be requiring a label.

tturocy (changes requested):

src/games/gameagg.cc
```diff
@@ -179,8 +179,8 @@ template class AGGMixedStrategyProfileRep<Rational>;
 GameAGGRep::GameAGGRep(std::shared_ptr<agg::AGG> p_aggPtr) : aggPtr(p_aggPtr)
 {
   for (int pl = 1; pl <= aggPtr->getNumPlayers(); pl++) {
-    m_players.push_back(std::make_shared<GamePlayerRep>(this, pl, aggPtr->getNumActions(pl - 1)));
-    m_players.back()->m_label = lexical_cast<std::string>(pl);
+    m_players.push_back(std::make_shared<GamePlayerRep>(this, pl, lexical_cast<std::string>(pl),
```
tturocy, line 182: This is a good example of a place where having the ctors have a parallel interface forces us to write code better in other places!

src/games/gametree.cc
```diff
@@ -1524,10 +1524,11 @@ int GameTreeRep::BehavProfileLength() const
 //                        GameTreeRep: Players
 //------------------------------------------------------------------------
 
-GamePlayer GameTreeRep::NewPlayer()
+GamePlayer GameTreeRep::NewPlayer(const std::string &p_label)
 {
+  auto player = std::make_shared<GamePlayerRep>(this, m_players.size() + 1, "");
```
tturocy: This defeats the purpose of having the label as an argument of the ctor if we allow ourselves to create a player with an invalid label.

What should be done is the validity of the label is checked first, and then the object allocated after the validity is confirmed.

This means that the logic in `SetLabel` for duplicate checking needs to be its own function that gets called from wherever is relevant.

tturocy (changes requested):

tturocy (approved):

PR #981, opened 2026-07-08
Enforce unique, nonempty labels for strategies

d-kad:
### Issues closed by this PR
- Part of #861

Enforces label rules for strategies, following the same pattern as players. Strategies are a nonempty, unique-within-player scope, enforced in `GameStrategyRep::SetLabel`:
- A strategy label must be nonempty, valid, and unique among a player's strategies; a violating label raises `ValueError`.
- `add_strategy` now requires a label (no default, no auto-generation).

### DIFF
--- a/src/games/gametable.cc
+++ b/src/games/gametable.cc
@@ -531,9 +531,9 @@ GameStrategy GameTableRep::NewStrategy(const GamePlayer &p_player, const std::st
   if (p_player->GetGame().get() != this) {
     throw MismatchException();
   }
-  auto strategy =
-      std::make_shared<GameStrategyRep>(p_player.get(), p_player->m_strategies.size() + 1, "");
-  strategy->SetLabel(p_label);
+  p_player->CheckStrategyLabel(p_label);
+  auto strategy = std::make_shared<GameStrategyRep>(p_player.get(),
+                                                    p_player->m_strategies.size() + 1, p_label);
   IncrementVersion();
   std::vector<long> old_radices;
   for (const auto &player : m_players) {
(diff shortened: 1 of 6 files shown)


tturocy:
I've added a commit that ensures generation of a unique strategy label in the GUI.

However, I noticed duplicate strategy labels are *not* being checked on creating new strategies.

src/pygambit/game.pxi
```diff
@@ -2194,16 +2199,23 @@ class Game:
...
+            If `label` is empty or is already the label of another of the player's strategies.
         """
         if self.is_tree:
             raise UndefinedOperationError(
                 "Adding strategies is only applicable to games in strategic form"
             )
         resolved_player = cython.cast(Player,
                                       self._resolve_player(player, "add_strategy"))
-        label_bytes = (str(label) if label is not None else "").encode("ascii")
+        if not label:
```
tturocy: Similar to my comment on the PR for player labels - these checks are hiding the fact that in C++ you can add a strategy with a duplicate label, which is a bug.  We should be able just to rely on the C++ logic to raise exceptions (possibly re-packaging then in Python/Cython if appropriate).

tturocy (changes requested):

src/games/gametable.cc
```diff
@@ -531,13 +531,15 @@ GameStrategy GameTableRep::NewStrategy(const GamePlayer &p_player, const std::st
   if (p_player->GetGame().get() != this) {
     throw MismatchException();
   }
+  auto strategy =
```
tturocy: Same comment as made on the player PR - the point of having a label that is required on initialisation is to provide a valid one.

tturocy (changes requested):

tturocy (approved):

PR #987, opened 2026-07-10
Add EFGs from MAS

wyz2368:
1. Add EFGs from MAS and some original layouts.
2. Fix Rahul's comment about imperfect recall and overlapped information sets.


tturocy:
@rahulsavani What do you think about the slug for the textbook (`shohambrown`)?

rahulsavani:
@tturocy 

> @rahulsavani What do you think about the slug for the textbook (`shohambrown`)?

Good point -- I'd totally missed that and would prefer, e.g., shohamleytonbrown2008, even though it's much longer. What do you think? Any better ideas?

tturocy:
> @tturocy
> 
> > @rahulsavani What do you think about the slug for the textbook (`shohambrown`)?
> 
> Good point -- I'd totally missed that and would prefer, e.g., shohamleytonbrown2008, even though it's much longer. What do you think? Any better ideas?

Given our conventions so far that would be what we would use; as we build the catalog going towards 17.0 we might review this and think if there's some convention for abbreviations before we start getting inconveniently long canonical names.

rahulsavani:
> > @tturocy
> > > @rahulsavani What do you think about the slug for the textbook (`shohambrown`)?
> > 
> > 
> > Good point -- I'd totally missed that and would prefer, e.g., shohamleytonbrown2008, even though it's much longer. What do you think? Any better ideas?
> 
> Given our conventions so far that would be what we would use; as we build the catalog going towards 17.0 we might review this and think if there's some convention for abbreviations before we start getting inconveniently long canonical names.

@wyz2368 OK, please update the slug:

shohambrown -> shohamleytonbrown

wyz2368:
@tturocy @rahulsavani Updated.

rahulsavani (changes requested):
The "original layout" version of fig6_8 now looks good to me, thanks. The "default" version still has the overlappying infosets -- can you fix that one. I saw you changed gtdraw_settings.yaml, but it seems those settings do not fix this.

rahulsavani:
> The "original layout" version of fig6_8 now looks good to me, thanks. The "default" version still has the overlappying infosets -- can you fix that one. I saw you changed gtdraw_settings.yaml, but it seems those settings do not fix this.

`sublevel_scaling` is probably sufficient to fix this.

wyz2368:
@rahulsavani I forgot to update the slug for the fig6_8 setting. Now it should work. Please check.

Screenshot 2026-07-10 at 12 16 23 PM

Screenshot 2026-07-10 at 12 16 41 PM

rahulsavani (approved):
Looks good to me

PR #992, opened 2026-07-20
Add xfail tests documenting non-unique generated RSF labels

d-kad:

Reduced-strategy labels are generated by concatenating action numbers with no separator, 
so they are non-unique once an information set has ten or more actions:  assignments `(1, 11)` and `(11, 1)` 
both render as `"111"`. This violates the per-player label uniqueness that #981  enforces on user-set labels, 
and it breaks NFG round-tripping — `WriteNfgFile` emits the duplicate labels and `ReadNfgFile` then rejects.

This PR documents the bug with two `xfail` tests for the smallest tree that produces exactly one collision.


tturocy:
This is a good point and something that was in the back of my mind.  But also as I note in #991, actually we should be thinking about what we want auto-generated labels to be especially now that actions do (or shortly will) be required to have nonempty labels.

tturocy:
Closing this as we have decided the problem is with this way of generating strategy labels, period!

PR #999, opened 2026-07-21
Add make_infoset and rewrite information-set operations in terms of it

d-kad:
Adds `Game.make_infoset(nodes, player, label=None)`, which forms a set of personal decision nodes 
into a single information set. This is the primitive for editing information structure; 
`leave_infoset`, `set_infoset`, `set_player`, and `reveal` are rewritten in terms of it.

Behavior changes:
- `set_infoset` now requires the node's actions to match the target infoset's, with the same labels in the same order
   (previously only the count was checked), and rejects terminal and chance nodes.
- `reveal` now raises on an absent-minded information set (previously the result was undefined).

Adds tests for `make_infoset` and for the four rewritten operations, which previously had little or no coverage.

Implemented in Cython over existing C++ operations; a native C++ `MakeInfoset` and the choice of which derived operations to retain are the following steps.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -15,10 +15,23 @@
   removes dependency on tinyxml.  (#897)
 - Enumeration of candidate supports for a totally-mixed Nash equilibrium is now done by a
   lazy generator algorithm rather than materialising all as a single list up front.
+- Introduced `Game.make_infoset`, which subsumes operations previously written as `Game.leave_infoset`,
+  `Game.set_infoset`, and `Game.set_player`.
+- `Game.reveal` now raises at absent-minded information sets; previously the result was undefined.  (#999)
+
+### Added
+- Added `Game.make_infoset`, forming a collection of personal decision nodes into a single
+  information set; the primitive operation for editing information structures.  (#999)
 
 ### Fixed
 - Converting a behavior profile to a strategy profile, and computing pure-strategy payoffs
   no longer inserts spurious entries into a strategy's internal action map.
+
+### Removed
+- `Game.set_player` has been removed; use `Game.make_infoset`, specifying the members of the
+  information set and the new player.  (#999)
+- `Game.sort_infosets` has been removed; it was deprecated in 16.5.0, as the iteration order of
+  information sets and their members is now maintained automatically.  (#999)
 
 ## [16.7.0] - 2026-07-11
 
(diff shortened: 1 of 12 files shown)


tests/games.py
```diff
@@ -415,6 +415,24 @@ def create_one_shot_trust_efg(unique_NE_variant: bool = False) -> gbt.Game:
     return g
 
 
+def create_two_pair_infosets_efg() -> gbt.Game:
```
tturocy: We're trying to avoid introducing more games to this file - we want to be removing them actually.  Can this test be accomplished by another existing catalog game, or if not, we should make such functions utility functions located close to the test(s) they are relevant for.

tests/test_node.py
```diff
@@ -601,6 +601,30 @@ def test_node_leave_infoset():
     assert list(proxy.members) == [node]
 
 
+def test_leave_infoset_sole_member_is_noop():
```
tturocy: I would move these under `test_infosets`, as they are manipulations of information structure.

(Overall our organisation of tests by the type of object is not ideal anyway.)

ChangeLog
```diff
@@ -1,6 +1,15 @@
 ## [17.0.0] - unreleased
 
+### Added
+- Added `Game.make_infoset`, forming a collection of personal decision nodes into a single
+  information set; the primitive operation for editing information structures.
+
 ### Changed
+- `Game.leave_infoset`, `Game.set_infoset`, `Game.set_player`, and `Game.reveal` are now
```
tturocy: Implementation detail changes are probably not important.  Changes to user-visible behaviour are, but ideally would be individual bullet points as they're actually logically separate (but I agree doing them all at once by thinking about edge cases while reimplementing makes sense rather than proliferating PRs)

tturocy (changes requested):
I think this is mostly looking good.  Are there any places in our existing tutorials or other materials where we should be using `make_infoset` in place of the other functions we are defining in terms of `make_infoset`?

See the main PR thread for other higher-level questions/next steps.

tturocy:
The next action on this is to push the implementation of `MakeInfoset` into C++.  This will allow us to unblock removing the other infoset-related actions by calling `MakeInfoset` from the GUI.

A new complementary issue #1009 provides a proposed way forward of dealing with nodes which heretofore we have called "chance infosets".

tturocy (approved):

PR #1004, opened 2026-07-22
Extend automatic ChangeLog extractor to support pre-releases

### LINKED ISSUE #994: [MAINT]: Extend ChangeLog extractor to support full semantic versioning versions
Our utility script to extract ChangeLog entries only supports versions in the form `X.Y.Z`.  This should be extended to support richer Semantic Versioning strings - in particular we would like to support `X.Y.Z-alpha.W`, `X.Y.Z-beta.W` and `X.Y.Z-rc.W` per planned releases leading up to 17.0.

tturocy:


tturocy:
@edwardchalstrey1 We don't want to impose on your good offices - but in this case, as this is a script you'd whipped up, would you be willing to take a few moments to see if we have covered the angles in how this should work if we want to allow more general SemVer releases?

PR #1007, opened 2026-07-28
HP tests

AndresFerCervell:
This PR adds five new tests to our test suite. 
The games tested have been selected from the catalog, using `enumpoly_solve` to choose the ones with a single Nash equilibrium.

The expected result is the equilibrium computed by the `enumpoly` solver.
The new games used are:

-`books/myerson1991/fig2_1.efg`
-`books/vonstengel2022/fig10.1.efg`
-`journals/ijgt/nau2004/sec4.nfg`
-`journals/other/reiley2008/fig1.efg`
-`journals/ijgt/selten1975/fig2` which has a degenerate prior (player 1 can choose three strategies with payoff = 1 when `t=0`).
They all pass the test!
### How to review this PR
Executing `pytest` and checking that the expected results are actually the NE of these games.


AndresFerCervell:
This PR has become a little chaotic since I wanted to use the new added games from the catalog which were not in feature/hp yet. 
About the new commit: 
Using `2x2x2.nfg` as discussed, new tests have been added to `test_hp.py` that explore different priors. When choosing a prior that is exactly a pure equilibrium, or a small perturbation of it, the mentioned equilibrium is returned. If we select a prior that is a mixed equilibrium (actually a perturbation of it in order to avoid various best responses), also a pure equilibrium is returned. 

Furthermore, now `test_hp.py` runs every game from the catalog that has a unique best response. The result´s `max_regret` is compared with 1e-8. They all pass in Linux and Windows, but `books/vonstengel2022/fig10.12`ends with ` 6.228842828764902e-05 = max_regret()`. For now, I have removed it from the test suite, but we could discuss what to do with it.

As well as that, I have also tested every game with a unique best response (this has been computed by the algorithm since we have already checked that it works as expected) from `contrib/games`. I have added to the test suite only a brief selection of the ones I found most interesting so that the execution of `pytest` Is not eternal. I would be glad to add more in the future if is required. Here, the algorithm fails sometimes to reach a NE with the asked tolerance. To be exact, in the following games:
-`todd1`
-`cent2`
-`wink3`
-`g2`
-`coord3`
-`todd2`
-`yamamoto`
-`3x3x3`
The error always comes from the beginning, the tracer starts from `t=0`, but it selects the wrong direction and goes to negative numbers, often leading to an infinite loop. I have solved this issue by modifying `hp.cc`. In a few minutes a new PR will be available explaining the changes. After this modification, only `wink3` does not get to reach an equilibrium.  However, when changing `c_maxDist = 0.4`to 0.2, in `path.cc`, it does pass the test.

tturocy:
Mistakenly closed this, we do want to keep this open for the new tests even though other code changes are getting done in separate PRs

PR #1012, opened 2026-07-31
Add make_chance_event and remove set_chance_probs

### LINKED ISSUE #1009: [API]: Develop `make_chance_event`
Gambit has always supported "information sets" for the chance player, even though this is not a part of most (all?) standard definitions of extensive form games.  However the semantics of this have always been awkward, because chance is not a player (and therefore doesn't have "information" per se), and that chance moves require specified probability distributions.

We propose that we will retain this notion, but instead refer to these as **chance events**.  Internally we do not need to make any changes - we for the moment retain the same internal data structures.  What will change is the API terminology.

To this end we want to create `make_chance_event(nodes, probs[, label])`.  This is the parallel of the new `make_infoset` operation (see PR #999).  It will accept a collection of nodes all of which must have conformant action labels (but need not all currently be chance nodes).  It also requires a specification of the probability distribution `probs`.  This can be specified either as a list-like of numbers of the correct length (in which case they match up one-to-one with the actions in order), or a dict-like mapping action labels to numbers.  In the latter case the dict-like can be sparse (in which case omitted actions are given zero probability).  In both cases, probabilities must sum to one and all be nonnegative.  An optional label can be specified - the rules for this label are the same as for `make_infoset`. 

As a temporary bridge, `set_chance_probs` can be re-written then to call this new function.  I argue that likely we will simply remove `set_chance_probs` as this function is arguably a semantically more clear way to express a change of probabilities if/when required.

We do not need to worry about changing the other infoset methods to run through this, as we are getting rid of those anyway.

d-kad:
### Issues closed by this PR
- Closes #1009

This PR adds `Game.make_chance_event(nodes, probs[, label])` and removes `Game.set_chance_probs`.

`MakeChanceEvent` is implemented in C++ as the parallel of `MakeInfoset` (#999). It additionally requires 
a probability distribution, given either as a list matching the actions in order or as a possibly-sparse mapping 
from action labels to numbers. Nodes need not already be chance nodes; personal decision nodes are converted. 

`set_chance_probs` is removed, subsumed by `make_chance_event`.

Depends on `GamePlayerRep::CheckInfosetLabel`, extracted here from `GameInfosetRep::SetLabel` with no behaviour change; the same commit is on the #999 branch and will drop out on rebase.

### DIFF
--- a/src/games/game.h
+++ b/src/games/game.h
@@ -1426,11 +1426,11 @@ public:
   //@{
   /// Set the probability distribution of actions at a chance node
   virtual Game SetChanceProbs(const GameInfoset &, const Array<Number> &) = 0;
-  /// Form the collection of nodes into a single chance event carrying the given
+  /// Form the collection of nodes into a single event carrying the given
   /// probability distribution over its actions.  The nodes need not currently be
   /// chance nodes; personal decision nodes are converted.
-  virtual GameInfoset MakeChanceEvent(const std::vector<GameNode> &, const std::vector<Number> &,
-                                      const std::string &)
+  virtual GameInfoset MakeEvent(const std::vector<GameNode> &, const std::vector<Number> &,
+                                const std::string &)
   {
     throw UndefinedException();
   }
--- a/src/games/game.h
+++ b/src/games/game.h
@@ -23,11 +23,13 @@
 #ifndef LIBGAMBIT_GAME_H
 #define LIBGAMBIT_GAME_H
 
+#include <algorithm>
 #include <list>
+#include <memory>
+#include <numeric>
+#include <queue>
 #include <set>
 #include <stack>
-#include <queue>
-#include <memory>
 
 #include "number.h"
 #include "gameobject.h"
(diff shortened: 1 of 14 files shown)


src/games/game.h
```diff
@@ -666,7 +671,7 @@ inline GameNodeRep::Actions::iterator::iterator(GameInfosetRep::Actions::iterato
 
 inline GameNode GameNodeRep::Actions::iterator::GetOwner() const { return m_child_it.GetOwner(); }
 
-inline void ValidateDistribution(const Array<Number> &p_probs, const bool p_normalized = true)
+template <class C> void ValidateDistribution(const C &p_probs, const bool p_normalized = true)
```
tturocy: I would feel slightly more comfortable with explicit overloaded implementations rather than a template, given the comparison internally to rational and the machinery of the `Number` inside the implementation

src/pygambit/game.pxi
```diff
@@ -1972,6 +1950,77 @@ class Game:
...
+            unique among the game's chance events after the operation.
+        """
+        if not self.is_tree:
+            raise UndefinedOperationError(
+                "make_chance_event(): operation only defined for games with a tree representation"
+            )
+        resolved_nodes = self._resolve_nodes(nodes, "make_chance_event")
+        if any(n.is_terminal for n in resolved_nodes):
+            raise UndefinedOperationError(
+                "make_chance_event(): all nodes must be decision nodes"
```
tturocy: "decision" -> "nonterminal" ?

src/pygambit/game.pxi
```diff
@@ -1972,6 +1950,77 @@ class Game:
...
+            )
+        resolved_node = cython.cast(Node, resolved_nodes[0])
```
tturocy, line 2021: Need to guard against an empty selected set!

src/pygambit/game.pxi
```diff
@@ -1972,6 +1950,77 @@ class Game:
...
+        action_labels = [a.label for a in resolved_node.infoset.actions]
+        resolved_probs = self._resolve_probs(probs, action_labels, "make_chance_event")
```
tturocy: We're not actually checking the nodes start out with conformant action sets?

src/pygambit/game.pxi
```diff
@@ -1972,6 +1950,77 @@ class Game:
...
+        c_nodes = stdvector[c_GameNode]()
+        for n in resolved_nodes:
+            c_nodes.push_back(cython.cast(Node, n).node)
+        c_probs = stdvector[c_Number]()
+        for p in resolved_probs:
+            c_probs.push_back(_to_number(p))
+        self.game.deref().MakeChanceEvent(c_nodes, c_probs, (label or "").encode("ascii"))
```
tturocy: We have a separate PR to move to Unicode  - should just change this directly to match the encoding done there.

tests/test_infosets.py
```diff
@@ -104,6 +104,73 @@ def test_infoset_plays():
     assert set(test_infoset.plays) == expected_set_of_plays
 
 
+@pytest.mark.parametrize(
```
tturocy, line 113: Looking also across at `make_infoset`, I think there's two tests missing:

* Trying to re-use a nonempty label without fully subsuming an existing move with that label
* Raising exception on `not game.is_tree`

tturocy (changes requested):

tturocy (approved):

PR #1016, opened 2026-08-01
Addresses two sources of inconsistency in the output of `enumpoly`:

### LINKED ISSUE #591: System-dependent behaviour of enumpoly on 2x2x2.nfg
On the standard game `2x2x2.nfg` (the example from McKelvey-McLennan with nine equilibria, the maximum number possible in a game of that size), there are two totally mixed equilibria.

When running `gambit_enumpoly` on the full support, at the starting point of the search (the centroid), the system of equations has a singular Jacobian.

This was previously noted in for example #474.  The fix for that resolved missing one of the equilibria on Windows.

However we have observed that the profile order on Windows is sometimes different from on other platforms (see #589).  In the Github action on that issue, it was noted that the order on Windows does not seem to be deterministic - both totally-mixed equilibria are found, but the order is different on different runs.  As this is in principle a deterministic algorithm, this behaviour is surprising.

tturocy:
1.  In the polynomial system solver, checks whether perturbations attempting to move off of a singular point are successful, rather than failing silently.
2.  Make the iterator over sequences iterate sequences in a deterministic order.


tturocy:
@rahulsavani We can put this on our agenda for the next meeting, specifically w/r/t whether we want to do anything else with the tests for this method.

rahulsavani (commented):
This looks good, but I think we should expand the tests a bit. I will do that, but welcome some input:

- I will make the 4x4 coordination game not stop after only 1, but at least a few more
- I will add Bernhard's 75 eq 6x6 game, as an efg and again set a >1 but not too large stop after (by the way, finding all seems super slow); I will also take this opportunity to put that game in the catalog in a separate PR
- I will add some "true" EFGs with multiple equilibria, which will perhaps be the most important addition, maybe one with 2 players and one with 3 players

@tturocy 
For the last bullet, I can come up with some examples myself, but wonder if you know any such games from papers or textbooks that would be suitable and that I could put into the catalog at the same time (I will take a pass on the existing catalog too in case I forgot about any suitable ones)?

PR #1019, opened 2026-08-02
Refactoring HPsystem

AndresFerCervell:
Refactoring of `hpsystem.h` and `hpsystem.cc`.
Now they follow the structure of `nfglogit.cc`, where each kind of equation has its own class that inherits from `HPEquation`.

No bugs have been found while testing them with the `pytest` suite and the games from `contrib`.

### DIFF
--- a/src/solvers/hp/hpsystem.cc
+++ b/src/solvers/hp/hpsystem.cc
@@ -25,10 +25,146 @@
 #include "solvers/hp/hpsystem.h"
 
 namespace Gambit {
+
+class HPEquation {
+public:
+  virtual ~HPEquation() = default;
+
+  virtual double Value(const Vector<double> &point,
+                       const MixedStrategyProfile<double> &current_sigma,
+                       const std::vector<double> &payoffs_against_prior) const = 0;
+
+  virtual void Gradient(const Vector<double> &point,
+                        const MixedStrategyProfile<double> &current_sigma,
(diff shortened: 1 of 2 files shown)


src/solvers/hp/hpsystem.cc
```diff
@@ -25,131 +25,212 @@
...
-          }
-          alpha_col++;
-        }
-      }
-      eq_idx++;
-      flat_s1_idx++;
+void HPEquationSystem::GetJacobian(const Vector<double> &point, Matrix<double> &p_jac) const
+{
+  // Update internal mutable state
+  int temp_alpha_idx = 2;
```
tturocy: This loop is repeated between `GetValue` and `GetJacobian`; would be suitable for a small helper

src/solvers/hp/hpsystem.cc
```diff
@@ -217,7 +298,8 @@ HPEquationSystem::ExtractEquilibrium(const Vector<double> &final_point) const
   for (const auto &player : m_game->GetPlayers()) {
     for (const auto &strategy : player->GetStrategies()) {
       const double alpha_val = final_point[alpha_idx++];
-      const double prob = this->AlphaToSigma(alpha_val);
+      const double prob = AlphaToSigma(alpha_val);
+      ;
```
tturocy: Null statement!

tturocy (changes requested):

PR #1021, opened 2026-08-03
Support UTF-8 labels throughout.

### LINKED ISSUE #862: [ENH]: Support unicode for labels

tturocy:
This implements the planned standardisation on UTF-8 labels for all strings (object labels, game titles, game descriptions).

Encoding/decoding is now done consistently in `pygambit`.

There was inconsistency in the wxWidgets encoding/decoding (between using the system conversion and UTF-8); this has been standardised.

The temporary ASCII-only wxWidgets label editor has been removed.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,6 +1,12 @@
 ## [17.0.0] - unreleased
 
 ### Changed
+- Game object labels may now be any well-formed UTF-8 text, not just printable ASCII; a label must
+  still not contain control characters, begin/end with whitespace, or contain two consecutive
+  whitespace characters -- where "whitespace" means any Unicode space separator (e.g. U+00A0
+  NO-BREAK SPACE), not just the ASCII space.  A game's title and description have no such
+  printable-character or spacing restriction, but must also be well-formed UTF-8.  In `pygambit`,
+  labels/title/description are now encoded/decoded as UTF-8 rather than ASCII.  (#862)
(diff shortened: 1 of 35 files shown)


tturocy:
@d-kad as you were involved in the temporary restriction to ASCII, please look through this, especially vis-a-vis test cases.

d-kad (commented):
Looks good — built and ran the branch.

One observation, attaching screenshots: the spacing rules seem to apply only to U+0020, so a leading U+00A0 (Option+Space on macbook) isn't stripped by `LabelTextCtrl::Normalize` and is accepted by `IsValidLabel`.

Image
Image
Image

tturocy:
That is a good catch, I did not update the meaning of "whitespace", which I've now done with the latest push

d-kad:
@tturocy looks good to me!

Issue #990, opened 2026-07-15
[GUI]: Refactor control of long-running computational processes


tturocy:
Currently, the graphical interface makes calls to the external `gambit-*` binaries to compute equilibria.

We would like to accomplish the following:
* [ ] Refactor this to be more robust - we will want to continue to have this as a possibility as sometimes we must run external tools
* [ ] Develop a version of the controller that runs in a thread - this would be the preferred way to support "Gambit-native" implementations to avoid complications with finding the locations of tools, and so on.

PR #1022, opened 2026-08-04
Migrate some solvers to run in threads in GUI

tturocy:
This is a step towards #990 - our goal of managing long-running equilibrium computation in the GUI process rather than depending on external tools and running in external processes.

For this change we are focusing on methods which do not take starting points, and therefore are not embarrassingly parallelisable.  Migrating those requires a bit more thought and work and so will be done in a separate tranche.

The methods which now run in threads by default are:
* `enumpure`
* `enummixed`
* `lcp`
* `lp`
* `logit`

To test: Run on several games, but ideally especially on some examples which run long enough to exercise the cancellation button.  Cancellation is cooperative, so the solvers have to be instrumented to poll for a cancellation request periodically and then act to shut themselves down.  This attempts to wire this so polling happens acceptably often on games of a size that are sensible to load in the graphical interface.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -213,6 +213,7 @@ core_SOURCES = \
     src/core/core.h \
     src/core/util.h \
 	src/core/array.h \
+	src/core/cancel.h \
 	src/core/lazy.h \
 	src/core/vector.h \
 	src/core/segment.h \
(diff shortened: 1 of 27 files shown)


d-kad (approved):

d-kad:
the tests I did:

- enummixed for rand16 (16x16 random), all/strategic: stop responded immediately; closing the dialog mid-run likewise; GUI responsive throughout
- enummixed for rand10, all/strategic: completes instantly
- lcp for bayes1a.efg, all/extensive: matches `gambit-lcp`, rationals preserved
- lp for 2smp.efg, one/extensive: matches `gambit-lp`
- simpdiv for rand10: fallback intact

PR #1030, opened 2026-08-07
Adding a controlled direction in path.cc

AndresFerCervell:
This PR intends to update `path.cc ` so that the tracer always follows the curve in the direction chosen by the programmer. It has been proven useful in the new `HP`algorithm to avoid a random direction in `t=0`. `logit` has also been affected by the modifications: the headers have been updated.

The mechanism implemented is the following:

-In `path.h` an enum type named **`TraceDirection`**  has been declared. It can be `Positive` associated to the number 1, and `Negative`, (-1). Positive means that the tracer will move forward; negative backwards. 

-To control whether the tracer is headed into the right direction in the first iteration, it is required to know which is the index of the variable that we are following (`lambda`in `logit`, `t` in `HP`). Another parameter has been added to `TracePath` named **`tracking_index`**  that stores the position of that variable in the vector `x`.  Note that in `HP` it is the first position and in `logit` is the last one.

The check that we are heading the right way is as discussed: if the first step goes as expected (equivalent to the tangent being positive), the internal omega does not change. Otherwise, it is multiplied by -1.

An exception is raised if the tangent in the first step is 0.

The changes have succesfully passed `logit` and `hp` tests.

### DIFF
--- a/src/solvers/logit/path.cc
+++ b/src/solvers/logit/path.cc
@@ -128,8 +128,9 @@ void NewtonStep(Matrix<double> &q, Matrix<double> &b, Vector<double> &u, Vector<
 TracePathResult
 PathTracer::TracePath(std::function<void(const Vector<double> &, Vector<double> &)> p_function,
                       std::function<void(const Vector<double> &, Matrix<double> &)> p_jacobian,
-                      Vector<double> &x, double &p_omega, TerminationFunctionType p_terminate,
-                      CallbackFunctionType p_callback, CriterionFunctionType p_criterion,
+                      Vector<double> &x, TraceDirection p_direction, size_t tracking_index,
+                      TerminationFunctionType p_terminate, CallbackFunctionType p_callback,
+                      CriterionFunctionType p_criterion,
                       CriterionBracketFunctionType p_criterionBracket) const
 {
   const double c_tol = 1.0e-4;   // tolerance for corrector iteration
@@ -157,12 +158,26 @@ PathTracer::TracePath(std::function<void(const Vector<double> &, Vector<double> 
   QRDecomp(b, q);
   q.GetRow(q.NumRows(), t);
   p_callback(x);
+  bool first_step = true;
+  double p_omega = static_cast<int>(p_direction);
 
   while (!p_terminate(x)) {
     bool accept = true;
 
     if (fabs(h) <= c_hmin) {
       return {x, false, "Stepsize fell below minimum threshold."};
+    }
+
+    if (first_step) {
+      // Ensure that the tangent is oriented in the same direction as
+      // the path-following direction.
+      if (t[tracking_index] < 0.0) {
+        p_omega *= -1.0;
+      }
+      else if (t[tracking_index] == 0.0) {
+        return {x, false, "Initial tangent vector is orthogonal to path-following direction."};
+      }
+      first_step = false;
     }
 
     // Predictor step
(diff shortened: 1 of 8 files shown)


src/solvers/logit/path.h
```diff
@@ -80,10 +80,13 @@ class PathTracer {
   void SetStepsize(double p_hStart) { m_hStart = p_hStart; }
   double GetStepsize() const { return m_hStart; }
 
+  enum class TraceDirection { Positive = 1, Negative = -1 };
+
   TracePathResult
   TracePath(std::function<void(const Vector<double> &, Vector<double> &)> p_function,
             std::function<void(const Vector<double> &, Matrix<double> &)> p_jacobian,
-            Vector<double> &p_x, double &p_omega, TerminationFunctionType p_terminate,
+            Vector<double> &p_x, TraceDirection p_direction, size_t tracking_index,
```
tturocy: For notational consistency, should name the parameter `p_trackingIndex`

src/solvers/logit/path.cc
```diff
@@ -157,6 +158,8 @@ PathTracer::TracePath(std::function<void(const Vector<double> &, Vector<double>
   QRDecomp(b, q);
   q.GetRow(q.NumRows(), t);
   p_callback(x);
+  bool first_step = true;
+  double p_omega = static_cast<int>(p_direction);
```
tturocy: (1) Now that this is just a local variable, let's call it `omega` rather than using the `p_` prefix.

(2) The cast might be too cute for its own good (and also casting to an int and then assigning to a double is a bit odd).  It depends on the values of the enum being what they are.  It might be better to have this be an explicit conditional just so the intent is clear - that is the benefit of moving to the named enum rather than just relying on the sign of a number or a bool.

src/solvers/logit/path.cc
```diff
@@ -165,6 +168,18 @@ PathTracer::TracePath(std::function<void(const Vector<double> &, Vector<double>
       return {x, false, "Stepsize fell below minimum threshold."};
     }
 
+    if (first_step) {
+      // Ensure that the tangent is oriented in the same direction as
+      // the path-following direction.
+      if (t[tracking_index] < 0.0) {
```
tturocy: Should check `tracking_index` is not out of range.  If it is would be best to return a failure message rather than just letting the exception propagate out.

src/solvers/logit/path.h
```diff
@@ -80,10 +80,13 @@ class PathTracer {
   void SetStepsize(double p_hStart) { m_hStart = p_hStart; }
   double GetStepsize() const { return m_hStart; }
 
+  enum class TraceDirection { Positive = 1, Negative = -1 };
```
tturocy: Slightly better style is to put declarations like this at or close to the top of the class for better visibility.

src/solvers/logit/path.cc
```diff
@@ -165,6 +168,18 @@ PathTracer::TracePath(std::function<void(const Vector<double> &, Vector<double>
...
+        p_omega *= -1.0;
+      }
+      else if (t[tracking_index] == 0.0) {
```
tturocy: A float is almost never going to compare to exactly zero; even if we are at a case where the tangent in the tracking index is zero, what will come out of the numerics will probably be tiny in magnitude but not zero.

We should add a new named constant for the tolerance - I'd suggest setting it to 1e-8 in the first instance.

tturocy (changes requested):
This has come out really nicely, it's a very elegant improvement to the path following.  A few notes basically on style and edge cases.

PR #1032, opened 2026-08-08
Polish method

AndresFerCervell:
This PR introduces the new function **`PolishPoint`** (other name suggestions are welcome) in `path.cc`. Its goal is to reduce the regret of a point that is already close to an equilibrium found by the path-following algorithm. Fixing the value of a component of the point (t in HP), it uses a Newton-type method to find a nearby point with lower regret. The resolution of the system uses the already implemented methods `QRDecomp` and `NewtonStep`.

It returns a **`TracePathResult`** struct. Now, `TracePathResult` also stores the number of  **steps** necessary to finish the algorithm, useful for debugging purposes. 
`PolishPoint` is used in `hp.cc` to refine the equilibrium computed by the tracer.

The termination function has been updated: now it returns true if `t >= 1`.

The results are very positive. When adding the orientation fix from PR #1030, we see that now we can accurately control the regret of the outputs. 

For example, if we set `tol=1e-1` (the tolerance is extremely high, so the polish function does nothing), and test the games with a unique best response at `t=0` from `contrib/games`, the tests that do not reach a regret lower than `1e-14`  are the following: 
```
  - todd1.nfg -> Eq: 1, Regret: 8.71e-02, Max T: 1.2008
  - 5x4x3.nfg -> Eq: 1, Regret: 1.60e-02, Max T: 1.08448
  - cent2.nfg -> Eq: 1, Regret: 3.00e-02, Max T: 1.01178
  - g3.nfg -> Eq: 1, Regret: 9.42e-03, Max T: 1.01912
  - g1.nfg -> Eq: 1, Regret: 7.63e-03, Max T: 1.32698
  - mixdom2.nfg -> Eq: 1, Regret: 7.23e-03, Max T: 1.03742
  - 2x2const.nfg -> Eq: 1, Regret: 1.12e-02, Max T: 1.03507
  - g2.nfg -> Eq: 1, Regret: 5.73e-04, Max T: 1.35734
  - 2x2x2x2x2.nfg -> Eq: 1, Regret: 4.24e-03, Max T: 1.01367
```
(Eq means the number of equilibria found, but that is not of interest for this PR, nor is Max T).
If we set now `tol=1e-8`, the results are the following: 
```
  - 5x4x3.nfg -> Eq: 1, Regret: 3.95e-10, Max T: 1.08448
  - g3.nfg -> Eq: 1, Regret: 4.00e-10, Max T: 1.01912
  - g1.nfg -> Eq: 1, Regret: 5.15e-09, Max T: 1.32698
  - mixdom2.nfg -> Eq: 1, Regret: 3.91e-13, Max T: 1.03742
  - 2x2const.nfg -> Eq: 1, Regret: 5.17e-10, Max T: 1.03507
  - g2.nfg -> Eq: 1, Regret: 8.79e-14, Max T: 1.35734
  - 2x2x2x2x2.nfg -> Eq: 1, Regret: 5.39e-09, Max T: 1.01367
```
We can see that every failing game has a regret higher than the tol. 
If we decide to equal the tolerance of the tests with the one in `hp.cc`, they all pass!
Some of these games are in `test_hp.py` in PR #1007 as a representation.


src/solvers/hp/hp.cc
```diff
@@ -38,10 +38,20 @@ HPStrategySolve(const MixedStrategyProfile<double> &p_prior)
 
   const PathTracer tracer;
   double omega = 1.0;
+  const double t_target = 1.0;
+  const double tol = 1e-8;
 
-  auto termination_condition = [](const Vector<double> &point) { return point[1] >= 1.5; };
-  auto criterion_function = [](const Vector<double> &point,
-                               const Vector<double> &tangent) -> double { return point[1] - 1.0; };
+  auto termination_condition = [t_target](const Vector<double> &point) {
```
tturocy: We don't need to specify a termination condition at all now.  Actually by having both termination condition and criterion, the criterion never gets used.  The result is we stop once we cross t=1.  But that could be a ways past t=1, and possibly leave us with a profile that's far enough from Nash that our polishing step doesn't work.

Instead if we have no termination condition and let criterion do the work, we wind up somewhere really close to t=1.  Then the polish step will work (almost for sure all the time!)

src/solvers/logit/path.h
```diff
@@ -62,6 +62,7 @@ struct TracePathResult {
   Vector<double> final_point;
   bool status; // true if path tracing terminated successfully, false if it terminated due to error
   std::string message; // error message if status is false
+  int step;            // Step at which the tracing terminated
```
tturocy: Calling this `steps` feels better - yes it's the step number where it finished but also is the number of steps taken, and we use `steps` everywhere else

src/solvers/logit/path.cc
```diff
@@ -269,7 +272,74 @@ PathTracer::TracePath(std::function<void(const Vector<double> &, Vector<double>
...
+
+  while (!p_terminate(x)) {
+    if (steps >= max_iter) {
+      return {x, false, "Polishing exceeded maximum iterations.", steps};
+    }
+
+    p_function(x, y);
+    p_jacobian(x, jac_full);
+
+    // Square Matrix removing fixed_index column
```
tturocy: This comment is slightly misleading, and actually the comment on `jac_square` above already covers what this is

src/solvers/logit/path.h
```diff
@@ -92,6 +93,16 @@ class PathTracer {
   double m_maxDecel{1.1}, m_hStart{0.03};
 };
 
+// This function reduces the regret of a point that is close to an equilibrium that has been found
+// by the path-following algorithm.  Fixing the value of a component of the point, it uses a
+// Newton-type method to find a nearby point with lower regret.
+TracePathResult
```
tturocy: This feels like we're abusing the return type.  Even if the fields are the same, the idea behind these result return types is that they're semantically related to the method they're coming from.  So I'd suggest defining its own result type for this method.

src/solvers/hp/hp.cc
```diff
@@ -66,7 +76,30 @@ HPStrategySolve(const MixedStrategyProfile<double> &p_prior)
...
+        std::cout << "Regret: " << regret << " | Probabilities: ";
+
+        Vector<double> prob_vector = profile.GetProbVector();
+        std::cout << std::fixed << std::setprecision(5);
+        for (size_t i = 1; i <= prob_vector.size(); ++i) {
+          std::cout << prob_vector[i] << " ";
+        }
+        std::cout << std::endl;
+      });
+
```
tturocy, line 99: This doesn't actually check to confirm that the polishing was successful!

tturocy (changes requested):
The results on this polishing idea are indeed very promising.  A few adjustments to make, but I don't see anything fundamentally wrong with the approach or implementation!

src/solvers/hp/hp.cc
AndresFerCervell: I understand, but I see two problems with removing it completely:
 
-The tracer will return a False status even when it is successful, because the only way that it has now of finishing is by reducing the step size to the minimum.

-`vd.nfg` fails because the criterion calculations become unstable near `t=1`, so eventually it keeps growing to infinity. 

Therefore, I suggest an implementation of the termination function that makes use of the criterion, but also has a safeguard in case it fails, leaving the problem to the polishing function. 

It works by comparing the current t with the last one when `last_t` was bigger than 1. If `t > last_t`, criterion is not working properly, and we have to finish the tracer before t gets too far. 
It also ends if we have reached the expected equilibrium.

tturocy: Good points both.

For the first point, this is a case where the step size getting small means something different in the two "modes".  When we're in Newton mode, it's actually potentially a sign of success, not failure.  What we should do is add a check that when in Newton mode, when the step size is tiny, check the value of the criterion function against a constant tolerance - if within that tolerance return success rather than failure.

For the second point, that game's got continua of equilibria which makes it a very good example.  Here's the output from `enummixed`:

```
convex-1,1,0,0,0,0,1,0,0
convex-1,1,0,0,0,2/3,1/3,0,0
convex-2,0,0,1,0,0,0,1,0
convex-2,0,0,1,0,1,0,0,0
convex-3,0,0,3/4,1/4,0,0,1/4,3/4
convex-3,0,0,3/4,1/4,1/4,1/3,0,5/12
convex-3,0,0,3/4,1/4,1/4,0,0,3/4
convex-3,0,0,3/4,1/4,0,7/12,1/4,1/6
convex-4,0,0,0,1,0,0,0,1
convex-4,0,0,0,1,0,2/3,0,1/3
```

That is, the set of equilibria is the union of four convex sets (with the indicated extreme points).  Seeing where the method is converging to in relation to these sets would be useful.

This will therefore also be a good practical example to include in the writeup, because it's a concrete example of where actually running this on "real" games - as opposed simply to doing the math theoretically, or running on games with randomly-generated payoffs, which will never have such positive-dimension equilibrium sets!

AndresFerCervell: Interesting!
I just added that check.
I will add more tests with games like this when these branches get merged.

src/solvers/hp/hp.cc
```diff
@@ -40,10 +40,38 @@ HPStrategySolve(const MixedStrategyProfile<double> &p_prior)
...
+      }
+      else {
+        // Criterion function is not working; polish will do the job
+        if (t > last_t + tol) {
+          return true;
+        }
+      }
+    }
+    else {
+      // Criterion function might take t back to being minor than t_target
```
tturocy: "less than" - I can see "minor than" is a direct calque from Spanish and/or Catalan!

src/solvers/hp/hp.cc
tturocy: What I meant here is that we shouldn't use the result of the polishing step as an equilibrium if it failed.

src/solvers/logit/path.cc
```diff
@@ -132,14 +132,15 @@ PathTracer::TracePath(std::function<void(const Vector<double> &, Vector<double>
...
-  const int c_maxIter = 100;     // maximum iterations in corrector
+  const double c_tol = 1.0e-4;      // tolerance for corrector iteration
+  const double c_maxDist = 0.4;     // maximal distance to curve
+  const double c_maxContr = 0.6;    // maximal contraction rate in corrector
+  const double c_eta = 0.1;         // perturbation to avoid cancellation
+                                    // in calculating contraction rate
+  double h = m_hStart;              // initial stepsize
+  const double c_hmin = 1.0e-8;     // minimal stepsize
+  const int c_maxIter = 100;        // maximum iterations in corrector
+  const double newton_tol = 1.0e-8; // tolerance for Newton convergence
```
tturocy: Would be good to keep similar naming style - `c_newtonTol`.

src/solvers/logit/path.cc
```diff
@@ -164,7 +165,12 @@ PathTracer::TracePath(std::function<void(const Vector<double> &, Vector<double>
     bool accept = true;
 
     if (fabs(h) <= c_hmin) {
-      return {x, false, "Stepsize fell below minimum threshold.", steps};
+      if (newton && std::abs(p_criterion(x, t)) < newton_tol) {
+        return {x, true, "Path tracing terminated successfully (Newton convergence)", steps};
```
tturocy: Probably need something more user-interpretable here: "Path following terminated successfully at point satisfying criterion function."

src/solvers/logit/path.cc
```diff
@@ -229,7 +235,12 @@ PathTracer::TracePath(std::function<void(const Vector<double> &, Vector<double>
     if (!accept) {
       h /= m_maxDecel; // PC not accepted; change stepsize and retry
       if (fabs(h) <= c_hmin) {
-        return {x, false, "Stepsize fell below minimum threshold.", steps};
+        if (newton && std::abs(p_criterion(x, t)) < newton_tol) {
+          return {x, true, "Path tracing terminated successfully (Newton convergence)", steps};
```
tturocy: Same comment as above

src/solvers/hp/hp.cc
```diff
@@ -40,10 +40,38 @@ HPStrategySolve(const MixedStrategyProfile<double> &p_prior)
...
+      return true;
+    }
+
+    if (t > t_target) {
+      if (!has_crossed) {
+        has_crossed = true;
+      }
+      else {
+        // Criterion function is not working; polish will do the job
+        if (t > last_t + tol) {
```
tturocy: Don't need this level of nesting, can pull the `if` up a level

tturocy (changes requested):
A few minor adjustments noted but I think this is close once we clean them up.

Two actions:
1.   You'll need to resolve the merge conflicts against the new main `feature/hp` branch as this interacts with the improvements on setting direction merged a bit ago today
2.  Remember to clear out the print tracing instrumentation for the next (final) review.

tturocy (changes requested):
Should also have added

3.  Would it make sense to roll in the test games from that other PR into this now, assuming they are working well with this methodology?

AndresFerCervell:
I decided to add the tests in this PR to avoid conflicts merging the other one. They all pass, although I had to add the catalog games that were not available at the start of the project.

src/solvers/hp/hp.cc
```diff
@@ -86,6 +86,9 @@ HPStrategySolve(const MixedStrategyProfile<double> &p_prior)
       },
       x, direction, tracking_index, termination_condition, NullCallbackFunction,
       criterion_function);
+  if (!result.status) {
```
AndresFerCervell: This makes some of the complex tests fail because the Tracer is not able to reach such a strict tolerance. The error message is "Maximum iterations exceeded.". I think we should let the polish function finish the work even though the tracer failed.

tturocy: I noted some comments in a separate email but to keep things in one place:

The first example (fig6_8) actually the tracing is successful but incorrectly reports failure, basically because it doesn't have to Newton step at all to achieve the tolerance.  That's a simple adjustment.

The second example really needs some scrutiny to understand why it's failing so we can be precise about what this implementation computes!

AndresFerCervell:
Some ideas for tomorrow´s meeting: 
-As we discussed, I have updated the termination function and the condition for returning an empty list in case that the tracer fails. 
-I have tried 100 random games with different number of players and strategies, each with three different priors (default, default slightly perturbed and random). Out of 300, around 270 would pass. As suspected, the problem comes with the singularity of the Jacobian. Even though the fact that at `t=0` we don´t allow multiple best responses, as t advances, they appear and make the Jacobian singular (only in non-generic games. HP paper proves that can´t happen in generic ones). 
This problem can be mostly solved adding this after  `QRDecomp` in the tracer
```
    for(int i = 1; i < b.NumRows(); i++){
      if (std::abs(b(i,i)) < 1e-10){
        if(b(i,i) < 0)
          b(i,i) -= 1e-8;
        else
          b(i,i) += 1e-8;
      }
    }
```
After this, the tests from `pytest` pass, only 1/300 random game fails, and some of the constant sum games we talked about too. 

About the fastest step size and max acceleration combination, it seems to be around 0.03 and 1.5 respectively. I will write it with more detail in the third report, and in the outputs we´ll talk about tomorrow.

PR #1040, opened 2026-08-12
Implement `Game.set_actions` to declare the action list of an information set.

### LINKED ISSUE #378: ENH: Consolidate `add_action`/`delete_action` into single operation
## Overview

The add and delete actions operations could be coalesced into a single operation, which for the moment let's call `set_actions`. For a chance move, `set_actions(infoset, actions)` would require a mapping from actions to probabilities, where actions as usual could be specified by label or by action object (requiring the action to be an action at that information set). New actions could be added by introducing a new label, and actions deleted by dropping the label. This would also allow actions to be re-ordered - re-ordering the actions would result in the corresponding subtrees being reordered. For personal player moves, actions would just be a list instead of a mapping. This behaviour is similar to using `DataFrame.reindex` in pandas, where it's possible to add, delete, and reorder rows and columns arbitrarily.

Specifically, this was trigged by the realisation that `delete_action` on a chance information set necessarily involves dealing with the action probabilities.  This was addressed in #372 to renormalise them to sum to one, but it is thought this might be too implicit and lead to unexpected behaviour.

## Design

We propose a new operation `set_actions` (name is open for improvement), which would have Python signature

```python
set_actions(infoset, actions: Iterable[str], drop: bool = False, add: bool = True)
```

(and the straightforward corresponding C++ equivalent)

Behaviour:
* The collection of actions at the information becomes as specified.
* Actions which already exist at the information set are re-ordered according to the list.  Actions keep their corresponding subtrees (so children are likewise reordered)
* New actions can be added by including them in `actions`.  By default this is acceptable (the ``add`` parameter defaults to ``True``).  This can be made strict by setting ``add=False``, which would raise an exception if any label in `actions` is not already present at the information set.
* Actions can be deleted by failing to include them in `action`.  By default this is *not* acceptable (the ``drop`` parameter defaults to ``False``) and will raise an exception.  This can be relaxed by setting ``drop=True``.

## See also

* #372
* #303

d-kad:
### Issues closed by this PR
- Closes #378 

Adds `Game.set_actions(infoset, actions, probs=None, drop=False, add=True)`, declaring the full ordered
action list of an information set, matching by label.  Entries matching current labels keep those actions and
their subtrees, in the declared position, so listing the current labels in a new order reorders the actions; 
entries matching no current action create new actions with terminal stubs at every member;
current actions not declared are deleted along with their subtrees.  
Deletion is destructive, so `drop` must be set to confirm it.

`Game.add_action` and `Game.delete_action` are removed.  

At an event, the probability distribution is declared with the actions: `probs` takes a sequence or a mapping, 
exactly as `make_event` does, resolved against the declared `actions`, and sum to one.  A pure reordering
needs no `probs`, since each probability stays with its action.  Deleting an action therefore never renormalizes: the distribution over what remains is explicitly declared.

The last commit removes `DeleteAction` and `NormalizeChanceProbs` (its only caller) from the C++ core: after 
`Game.delete_action`'s removal neither has any caller in the tree.  
@tturocy let me know if I should drop this last commit if you would rather keep them.


tturocy (changes requested):
This is mostly looking OK, but we need to enhance the test coverage - specifically missing would seem to be permutations of more than two actions, or moving actions - and specifically making sure that when actions are moved the whole subtree underneath them goes with it (at the moment looks like only actions above terminal nodes are tried?)

tturocy:
@rahulsavani @d-kad I have drafted a new "Edit move" dialog which allows the exercise of the whole of `SetActions` and `RelabelActions`.  As a result `InsertAction` has been removed.  Meanwhile there was no way to delete an action in the GUI (possible in the underlying code but at some point got lost), and also it's never been possible to reorder actions in a single option without some complex dancing of copying subtrees manually.

Please have a go and comment as it surely can be improved, though I think it's a big step up from what we had!

d-kad:
I tested the edit move dialog, mostly good, the only bug I found is that reassigning the player raises "Dereferencing an invalidated object"

Screenshot 2026-08-17 at 14 28 47

Screenshot 2026-08-17 at 14 28 53

tturocy (approved):

PR #1050, opened 2026-08-16
Complete migration of GUI to run algorithms in threads

### LINKED ISSUE #990: [GUI]: Refactor control of long-running computational processes

tturocy:
This completes the migration of the GUI calls to use threads to run algorithms, by implementing the needed cancel-check points inside the remaining methods.

Closes #990.


tturocy:
@d-kad @rahulsavani As before, exercise the GUI here on the rest of these methods!

d-kad:
The tests I did:

* simpdiv for rand10, one eq: completes instantly
* ipa for rand10, some eq: completes instantly
* gnm for rand10, some eq: 5 profiles, including the pure (1:2, 2:5)
* enumpoly for rand3, one eq: matches `gambit-enumpoly`
* enumpoly for rand20, some eq: found 2 before I stopped -- responded immediately; GUI responsive throughout
* liap for rand6 and rand10, some eq: finds nothing, but `gambit-liap -n 10` found nothing either

d-kad (approved):

PR #1052, opened 2026-08-17
C++ api and internal cleanups

tturocy:
This is an omnibus of several commits focused on refactorings ultimately targeted at being able to remove direct reference to internal game objects (mediated via shared pointers)


tturocy:
Going to attack this from a slightly different angle.

PR #1056, opened 2026-08-18
Implement `Game.set_strategies` to declare the strategy list of a player.

d-kad:

Adds `Game.set_strategies(player, strategies, drop=False, add=True)`, which declares the ordered
list of a player's strategies in a game in strategic form, matching by label.  An entry matching a
current label keeps that strategy and the outcomes at its contingencies; an entry matching none
creates a strategy with no outcome at any of its contingencies; a current strategy whose label is
not listed is deleted, along with the outcomes at its contingencies.  Listing the current labels in
a new order reorders the strategies, permuting the payoff table to match -- not previously
expressible.  Deleting is destructive, so it must be confirmed with `drop=True`.

`Game.add_strategy` and `Game.delete_strategy` are removed, as `set_strategies` subsumes both.


tturocy:
Generally looking good.  @rahulsavani @d-kad the usual drill - left-clicking on strategies in the normal form GUI now generates a dialog with should operate just like the edit-move dialog does for actions.  Seems to work well for me, please put it through its paces!

d-kad:
Tested Edit strategies dialog on macOS: renaming, reordering, adding, deleting with restore all behave well.

The only problem I spotted: with a new 2x2 strategic game, adding players repeatedly crashes loudly at around the 5th player (screenshot)

Screenshot 2026-08-19 at 17 07 43

tturocy:
> Tested Edit strategies dialog on macOS: renaming, reordering, adding, deleting with restore all behave well.
> 
> The only problem I spotted: with a new 2x2 strategic game, adding players repeatedly crashes loudly at around the 5th player (screenshot)
> 
> Screenshot 2026-08-19 at 17 07 43

This turned out to be a bug in NewPlayer, completely separate from this.  I've patched it for this PR, but actually it should get fixed "automatically" in the `set_players` changes.

PR #1057, opened 2026-08-19
Improve numerical stability and robustness of `enumpoly`

### LINKED ISSUE #1055: [Bug]: enumpoly hanging on 6x6 bimatrix game

PyGambit

macOS

Development version

### What happened?

The bug is shown in the following example code where we load a 6x6 bimatrix game with 75 equilibria from the catalog.

Using `enummixed_solve` we find all 75 equilibria, but in the below code with `enumpoly_solve`, after creating an EFG version of the bimatrix game we then struggle to find the:

- 54 equilibria with `use_strategic=True` (but finding 53 is fine);
- 1 equilibrium with with `use_strategic=False`.

We find the same with the "small payoff" version of this game (slug: "6x6_game_with_75_eq_small_payoffs").

```
import pygambit as gbt
from tests.games import create_efg_corresponding_to_bimatrix_game  

slug = "journals/dcg/vonstengel1999/6x6_game_with_75_eq"
A, B = gbt.catalog.load(slug).to_arrays()
g = create_efg_corresponding_to_bimatrix_game(A, B, "")

#######################
# use_strategic is True
#######################

# Runs in ~ 10 seconds:
print(gbt.nash.enumpoly_solve(g, stop_after=53, use_strategic=True))

# Hangs:
# print(gbt.nash.enumpoly_solve(g, stop_after=54, use_strategic=True))

########################
# use_strategic is False
########################

# Hangs:
# print(gbt.nash.enumpoly_solve(g, stop_after=1, use_strategic=False))
```

tturocy:
This makes two improvements:

1. Adds a `max_rectangles` argument to the branch-and-bound equation solver, to prevent it from futilely subdividing down to rectangles of size 1e-8 everywhere;
2.  Rewrite the matrix inversion and solving in a dedicated LU decomposition object which should substantially improve reliability on poorly-condition systems.

Closes #1055.


rahulsavani (approved):
Very happy with this.

I added the tests I had in mind for `enumpoly`. It runs too slow under behavior strategies to do too many equilibria, so only 2 there, and 10 for the original strategic form version.

One thing I recalled by accidentally setting `stop_after` to 0 is that it is equivalent to `stop_after` being `None`. I think that `None` corresponding to "all" equilibria is natural, but for 0 much less so, so I wonder if that should be mentioned in the docs for any methods that have `stop_after` as a param? If you agree I can create a separate issue and do that shortly.

tturocy:
stop_after = 0 working is accidental leakage from the convention in the C++ code.  Please make an issue and we can discuss fixing it.

PR #1077, opened 2026-08-24
GUI: Add sorting functionality for equilibria

ameliekleber:

This PR makes the list of computed equilibria sortable, in both the extensive-form (behavior profile) and strategic-form (mixed profile) panels. Previously equilibria could only be read in the order the algorithm produced them, which makes them hard to compare or locate in a long list.

Clicking a column label sorts the equilibria by the probability with which that action or strategy is played; clicking again reverses the direction. Ties are broken by comparing profiles entry-by-entry from the leftmost column, so sorting on the first column lists them lexicographically by profile. Clicking the # corner restores the computed order.

Sorting only changes the display order: the row number identifies the equilibrium itself and stays with it, so selection and the bold highlight follow the profile, not the row.

For discoverability, each column label carries a ▲/▼ on the sorted column; the label windows also use a hand cursor and carry tooltips. In addition, this PR makes the permanent light blue highlight (on the first cell in the second column and the first cell of the second row) dynamic to emphasize which column is used for sorting, and which row is bolded.

The ordering lives in a new ProfileSortOrder (src/gui/profilesort.{h,cc}) shared by both panels. It compares numeric values rather than the formatted cell strings, so 1/3 and 0.333 order correctly (that is what the new GetStrategyProbValue/GetActionProbValue accessors on AnalysisOutput are for). Entries undefined for a profile (shown as *) sort last. doc/gui.nash.rst is updated to match.

### How to review this PR

- Rebuild the GUI from this branch, then test that the sorting works as expected


src/gui/analysis.cc
```diff
@@ -346,6 +346,36 @@ std::string AnalysisProfileList<T>::GetStrategyValue(int p_strategy, int p_index
   }
 }
 
+template <class T>
+std::optional<double> AnalysisProfileList<T>::GetActionProbValue(int p_action, int p_index) const
```
tturocy: Is there a rationale for casting to `double` even when the type is `T=Rational`?  Rationals should sort just fine as well.

doc/gui.nash.rst
```diff
@@ -149,6 +149,25 @@ the screenshot, it is set to
 brief description of the method used to compute the equilibria is
 listed across the top of the profiles panel.
 
+By default the equilibria are listed in the order in which they were
+computed. Whenever more than one equilibrium has been found, each
+column label in the profiles listing carries a small
+:guilabel:`↕` marker, indicating that the list can be sorted on that
```
tturocy: I found the this marker to be very difficult to see on my screen or understand what it means.  Can we just do without it?  I've seen many interfaces where we don't explicitly mark "hey you can sort this".

tturocy (changes requested):

src/gui/analysis.cc
ameliekleber: I think it was because this way it could compare cached keys, but now it calls back per comparison. However, I think for sorting a few equilibria, this should be fine, right?
I have made the changes!

tturocy (approved):

PR #1078, opened 2026-08-24
Remove `Action` object from `pygambit`

tturocy:
This removes `Action` from `pygambit` (as a direct wrapper holding a shared pointer to internal Game objects).

Actions are now principally referred to simply by their labels, as they are required to be nonempty and unique within their scope (information set/event).

Where required for context, an `Action` or `Branch` named tuple is provided (to link the action label to its corresponding information set/event or node, respectively).


tturocy:
@rahulsavani @d-kad For discussion and next action - these changes break `gtdraw` so they break tutorials - we need to action an update to `gtdraw` to use the new API.

tturocy:
This has been accomplished by another line of work.

PR #1113, opened 2026-09-03
Add Herings-Peeters homotopy method

tturocy:


tturocy:
@AndresFerCervell In this branch we've added in all of the instrumentation to be able to use this as an equilibrium solver (callbacks, ability to cancel in a multi-threaded computation), and hooked this up to a command-line driver and added it to the GUI.

There are two last things to do:
1.  Settle on what the "best" solution (for now) is for termination/polishing.  As discussed, let's go for something that maximises the chance we get an epsilon-Nash equilibrium for now, and we can investigate more carefully some of the difficult cases separately.
2. We'll need to finalise a section in the algorithms documentation which briefly describes the method.

AndresFerCervell:
They are handled in PR #1121 and PR #1120, respectively.

PR #1121, opened 2026-09-05
HP provisional termination

AndresFerCervell:
This PR solves the first part of PR #1113. It  includes the changes made in PR #1032, removing conflicts.

For every generic game, the Nash equilibrium selected by the Harsanyi Selten procedure is guaranteed (at least we haven´t found a counterexample).
Regarding non-generic ones, if an equilibrium is returned, there is no guarantee that it is the one we wanted (tracer result should be checked in order to be sure).
However, I haven´t  found any example in the 477 games evaluated (random and the ones from the catalog or contrib) that happen to find an equilibrium after a tracer error. This means that the polishing method is not being used (if the tracer fails, so does it).

The games that always fail are, as I mentioned in our meeting, `csg2.nfg`, `csg3.nfg` and `csg4.nfg` and a random game of 2 players with 10 strategies each. This one is specially confusing because the Jacobian does not seem to be singular. The tracer gets a regret of 0.00542893, and the polishing method returns an all-zero profile in every position in spite of the safeguard I added. I will keep studying it.

We can say then that around 470/477 games get the expected equilibrium.


AndresFerCervell:
The tests for this PR are failing on Python 3.10 with:
ERROR build_support/catalog/test_update.py - AttributeError: module 'pygambit.gambit' has no attribute 'Selector'
I guess it has something to do with the recent update of Gambit, but has no relation with my new code. Should I do something about it?

AndresFerCervell:
Just understood the issue with the random game. When it approaches to 1, the first position of the tangent is almost 0 in the newton step. Therefore, `t` barely moves, and, when computing the next step size `h`, `p_criterion(u, newT) - p_criterion(x, t)` is zero. 
We divide by that to get `h`, so it went to infinity. I have added a check that confirms the difference is greater than `1e-8`. 

The polishing method returned the all-zero profiles because the comparison between numbers and nan is false. Fixed that too.

AndresFerCervell:
Regarding csg, just found out that, at some point, the Jacobian drops to rank N-2. 

The perturbation that I added to the Jacobian is enough to solve the games only when they have a jacobian with N-1 rank. 

I have studied, as you suggested, the Givens Rotations and the QR decomposition. They actually seem to be working just fine, (at least in these cases). I don´t think that  swapping rotations for some more modern system such as Householder reflections would solve the problem. While it is true that Givens Rotations is not the standard in the industry right now for efficiency reasons, I don´t believe it´s worth it to change it now.   Adding a pivoting system in the already existing method would be easier and could make the refactor more robust. But again, I don´t think that would solve our problem.

PR #1133, opened 2026-09-10
Package catalog as part of GUI build and load games from catalog

tturocy:


rahulsavani (commented):
This works well for me, great stuff.

One idea: We might want to allow sorting by column in the catalog table, which would e.g., give an easy way, without a search function, to see all 3 player games together, or find a particular entry.

PR #1136, opened 2026-09-10
Expose enummixed's connected components as NashComponent/NashPolytope

tturocy:
This adds NashPolytope (one maximal clique, genuinely convex) and NashComponent (one or more NashPolytopes glued at a shared vertex, not itself necessarily convex) as view objects, with the grouping computed via a small union-find pass over the existing per-clique vertex-index arrays in EnumMixedStrategySolution. EnumMixedResult.cliques becomes .components, and enummixed_solve's cliques= parameter becomes components=.


tturocy:
@rahulsavani Here's a marker on this idea.  To be discussed/considered:
1. Whether a hit-test/containment operation should be added to the new polytope objects;
2. How might we present this in the GUI?

PR #1153, opened 2026-09-18
Draft update to `make_event`

### LINKED ISSUE #1147: `make_event` should require all probability distributions to be the same
At present `make_event` allows probability distributions to be specified (and worse yet, sparsely specified).   This is a violation of orthogonality and doesn't parallel `make_infoset` which requires the action lists already to be conformant.

Revise `make_event` so that it checks that all the histories specified in fact have the same actions and probability distribution prior to acting.

tturocy:
This closes the inconsistency in `make_event` (which allowed setting of event probabilities which was not what was intended).


tturocy:
@rahulsavani @wyz2368 This is a first step towards one of the items found in the survey review.  However, it leaves us with a problem: We do not have a way of converting from an information set (player move) to an event.

Within the current API as designed, I'm not sure where the best place for that to sit is in terms of operations.  But it is definitely an operation we want as there are clear use cases - transforming a game by assuming one of the players is actually nonstrategic and plays a fixed behaviour.

